Issue #27, opened 2017-03-31
Create an extensible polling solution


spmcbride1201:
Currently, the dash is using a bare-bones solution for polling (setInterval via lifecycle hooks on the parent React component). This works for being able to stand up a handful of visualizations as part of an MVP, but going forward, this application will have to ingest data from a variety of endpoints. More advanced use cases, like actually implementing the admin polling UI controls I've built have a soft dependency on this as well.

While there are alternative callback based approaches to async, I believe that generators are vastly more comprehensible to those new to asynchronous development, leading to more easily maintainable code.

spmcbride1201:
I plan to do this as part of a general refactor of the app to use Jumpstate.

PR #28, opened 2017-03-31
Implement basic Settings Component and fix broken chart

### LINKED ISSUE #10: Create Admin Component
This should initially allow the dashboard to customize the URL and port where metrics.json is being served.

### LINKED ISSUE #19: On Initial Load, Data Transfer graph kbps up and down values are NaN

spmcbride1201:
I've implemented two settings cards on a SettingsGrid component. The button clearing the Redux store is functional, but the settings for adjusting polling settings only dispatch redux actions that do nothing (and are labeled as such). Actually wiring these up are contingent on #27.

I've also made the DataTx chart a bit smarter in its Array.prototype.map helper function. Given that the logic is now a bit more complicated, I added some comments to explain the branching logic.
Resolves #10 
Resolves #19


tilleryd:
LGTM

PR #31, opened 2017-04-03
Refactored store using Jumpstate

### LINKED ISSUE #17: Refactor to use ES7 static property initializers
image

### LINKED ISSUE #27: Create an extensible polling solution

spmcbride1201:
Also cleans up code, wires up settings controls, and implements static property initializers

Resolves #27 
Resolves #17


.vscode/cSpell.json
```diff
@@ -0,0 +1,22 @@
+// cSpell Settings
+{
+    // Version of the setting file.  Always 0.1
```
tilleryd: Hrmm. Comments in the JSON aren't valid.

src/components/Container.js
```diff
@@ -43,12 +39,4 @@ class Container extends Component {
...
 
-Container.propTypes = {
-  fetchMetrics: PropTypes.func.isRequired
-};
-
-function mapDispatchToProps(dispatch) {
-  return bindActionCreators({ fetchMetrics }, dispatch);
-}
-
-export default connect(null, mapDispatchToProps)(Container);
```
tilleryd: Wow, seeing what's removed really demonstrates how JumpState greatly reduces Redux boilerplate.

src/store.js
```diff
@@ -1,14 +1,112 @@
...
+// Stop the interval, change the endpoint, and restart when state.settings.interval is changed while polling is live
+Hook((action, getState) => {
+  if (action.type === 'setInterval' && getState().settings.setMetricsEndpoint) {
+    Actions.stopPolling();
+    Actions.startPolling({
+      endpoint: getState().settings.metricsEndpoint,
+      interval: getState().settings.interval
+    });
+  }
+});
```
tilleryd: I think it's ok for now to have everything in 1 file.  As hooks and effects grow we could move we can move them to their own files.  It doesn't seem like this app will require individual files for each hook or effect so 1 file for hooks and 1 for effects should be fine.

.vscode/cSpell.json
tilleryd: This is the only thing from this review that should be changed.

tilleryd: Actually if this works it isn't a big deal since it's just being used in vsCode.

PR #35, opened 2017-04-05
Refactor Data Schema, implement utility functions, build out new charts, and merge sidebars.

### LINKED ISSUE #8: Use charting library to implement data visualizations

### LINKED ISSUE #26: Use normalizr to normalize twitter-server metrics.json
Currently, the dashboard is ingesting metrics into the Redux store as-is, which is forcing React components to perform somewhat complex mapping operations in the render loop. It also forces all of the metrics data to be passed en-masse as props.

Normalizr would allow us to structure the data in a way better suited for data visualization. A possible scheme might be as follows:
endpoint --> attribute --> timestamp --> value

Considering that the twitter-server metrics also have / delimited keys that convey a hierarchy of data, there are some opportunities to exploit this as well to organize and decompose the Redux store. 

Also worth consideration is use of `humps` to covert keys from underscores the camelCase.

### LINKED ISSUE #30: Use a common nav component between the oncanvas and offcanvas sidesbars

### LINKED ISSUE #37: When changing interval, the metrics sampled are not properly normalized in per second calculations
Currently, a utility function takes the difference between two snapshots of a single metric and calculates the rate of change by dividing by the interval between polling the server. However, the interval may have changed since last polling the server. To correct this, the calculation should divide by the difference of the timestamps of the two snapshots.

### LINKED ISSUE #38: Add custom favicon.ico
Presumably, this should be the gray matter logo

spmcbride1201:
Resolves #8.
Resolves #26.
Resolves #30.
Resolves #38.
Resolves #37.


src/components/DataTxLineChart.js
```diff
@@ -1,52 +1,17 @@
 import React, { PropTypes } from 'react';
-import dateFormat from 'dateformat';
 import { LineChart, Line, CartesianGrid, XAxis, YAxis, Tooltip, ResponsiveContainer } from 'recharts';
 
 DataTxLineChart.propTypes = {
-  metricsArr: PropTypes.array.isRequired
+  receivedAndSentBytesPerSecondArr: PropTypes.array,
```
tilleryd: trailing comma... no big deal but please remove

tilleryd: trailing comma... no big deal but please remove

src/components/RouteGrid.js
```diff
@@ -0,0 +1,55 @@
...
+          className="uk-grid-match uk-grid-small uk-text-center"
+          data-uk-grid
+        >
+          {routeVerbs.map(verbName => (
+            <div
+              className="uk-width-1-2@m"
+              key={verbName}
+            >
+              <div className="uk-card uk-card-default uk-card-body">
+                <h3 className="uk-card-title">{`${verbName} /${this.props.params.routeName}`}</h3>
```
tilleryd: We tend to use object destructuring to cut down on referencing props via `this.props`.

```
const { params } = this.props;
...
<h3 className="uk-card-title">{`${verbName} /${params.routeName}`}</h3>
```

This is more of a style thing.  I prefer not to create intermediate vars but it does make the code a bit more terse and easier to read.  No action necessary, just sharing our style.

src/index.js
```diff
@@ -11,15 +11,20 @@ import HTTPGrid from './components/HTTPGrid';
...
+import RouteGrid from './components/RouteGrid';
+import { syncHistoryWithStore } from 'react-router-redux';
 import '../node_modules/uikit/dist/css/uikit.min.css';
 import './style/index.css';
 
 // load the UIKit Icon plugin
 UIkit.use(Icons);
 
+// Wire up React Router with Redux
+const history = syncHistoryWithStore(browserHistory, store);
```
tilleryd: I recommend that you implement history like Up2.  We ran into an issue where custom baseUrls would break routing.  We resolved this as follows:

1. A gulp task sets the baseUrl in index.html
https://github.com/DecipherNow/up2-ui/blob/develop/gulpTasks/baseUrl.js
https://github.com/DecipherNow/up2-ui/blob/develop/src/index.html
2. We create router history using a custom `baseName` so history runs relative to the baseUrl
https://github.com/DecipherNow/up2-ui/blob/develop/src/js/store.js

tilleryd: This certainly is not a requirement of this PR and can be an issue for future work.

src/style/index.css
```diff
@@ -1,38 +1,44 @@
...
-  color: #ddd
+/*These Rules are to ensure that the sidebar can display as white */
+/*uk-offcanvas-close>svg {
+  color: #999 !important;
 }
 
-.hamburger:hover {
-  color: #fff
+.uk-nav>li>a, .uk-nav-sub>li>a {
+    color: #999 !important;
```
tilleryd: indentation

src/components/RouteGrid.js
spmcbride1201: Done! I think it probably did improve readability.

src/index.js
spmcbride1201: Cool. I opened Issue #40 for this.

PR #46, opened 2017-04-10
feat: Fallback to hashHistory and support baseurls

### LINKED ISSUE #14: Dynamically use HashHistory on older browsers
From @tilleryd:
so we're doing this actually
 ```// manually create history using a custom basename configuration, this will ensure
// that all routes will be run relative to the value of BASE_URL
const history = useRouterHistory(
  supportsHistory() ? createHistory : createHashHistory
)({
  basename: BASE_URL
});
```
take a look at https://github.com/DecipherNow/up2-ui/blob/develop/src/js/store.js

### LINKED ISSUE #40: Support Custom Base URLS for front-end routing
@tilleryd suggests implementing history like Up2. We ran into an issue where custom baseUrls would break routing. We resolved this as follows:

A gulp task sets the baseUrl in index.html
https://github.com/DecipherNow/up2-ui/blob/develop/gulpTasks/baseUrl.js
https://github.com/DecipherNow/up2-ui/blob/develop/src/index.html
We create router history using a custom baseName so history runs relative to the baseUrl
https://github.com/DecipherNow/up2-ui/blob/develop/src/js/store.js

spmcbride1201:
Resolves #14.
Resolves #40.


tilleryd:
Works as intended.  I tested on `master` with a baseUrl of `foo` and also with this `basename` branch.  It didn't work on `master` but did work on this branch as expected.

PR #69, opened 2017-04-21
Build Master and save to CircleCI as tarball

spmcbride1201:


tilleryd:
Do all these commits really pertain to the work at hand?

spmcbride1201:
No, but it works. You can review the other PR instead, and we can put this feature branch on ice until after Amsterdam

PR #85, opened 2017-06-30
Add experimental Golang support

spmcbride1201:
This PR merges in basic functionality for experimental golang support. It provides a single moving sparkline chart and a working JSON component to allow for further iteration on golang parsing.


src/utils.js
```diff
@@ -155,7 +156,19 @@ export function mergeResults(...arrayOfResultArrays) {
 // Reselect Selectors
 
 // Reselect Input Selectors
+
+/**
+ * Reselect Input selector that returns the threadsFilter from state.settings.threadsFilter
+ * Currently used by getVisibleThreads
```
tilleryd: Also used by `getThreadCounts`, do we really want to list which functions call this function?  I think we'll forget to maintain the list.

src/utils.js
```diff
@@ -193,8 +206,117 @@ export const getThreadCounts = createSelector(getCurrentThreads, (threadsTable =
...
+    }
+    else {
+      // Change from slash delimited to dot delimited and from snake case to camel case to be able to
+      // use the lodash set method to build a idiomatic JSON hierarchy of data
+      let dotDelimitedCamelizedPath = camelize(slashDelimitedPath.replace(/\//gi, '.'));
+      _.setWith(normalizedTimeseriesData, `${dotDelimitedCamelizedPath}.${timestampOfMetricsPoll}`, metricsValue);
+    }
+  });
+  return normalizedTimeseriesData;
+}
```
tilleryd: The JSDoc comments are really helpful.  Thank you.

src/utils.js
```diff
@@ -193,8 +206,117 @@ export const getThreadCounts = createSelector(getCurrentThreads, (threadsTable =
...
   const baseUrl = metaBaseUrl.indexOf('__BASE_') !== -1 ? "/" : `${metaBaseUrl}`;
   return baseUrl;
 };
+
+/**
+ * getRuntime is a utility function that extracts the runtime from the HEAD of the index.html file. It's used to determine
+ * how the app should scrape metrics and render React components
+ */
+export function getRuntime() {
+  const metaRuntime = document.head.querySelector("[property=runtime]").content;
```
tilleryd: I know we discussed that at some future point we could build the dashboard static assets and add them the respective MicroService starter kit.  If we do that then can't we just set the endpoints we need in a config file?  Then we wouldn't event need to detect the runtime as we would already have the endpoints.

src/utils.js
tilleryd: It appears you're missing return param types in some of your JSDocs.

src/utils.js
```diff
@@ -193,8 +206,117 @@ export const getThreadCounts = createSelector(getCurrentThreads, (threadsTable =
...
+ * form expected by the React UI components, and restructures as timeseries data using the UNIX timestamp of when the
+ * endpoints were polled.
+ * @param {Object} rawScrapedMetrics
+ */
+export function parseGolangMetrics(rawScrapedMetrics) {
+  let normalizedTimeseriesData = {};
+  let timestampOfMetricsPoll = Date.now();
+  // Change from slash delimited to dot delimited and from snake case to camel case to be able to
+  // use the lodash set method to build a idiomatic JSON hierarchy of data
+  _.forIn(rawScrapedMetrics, (metricsValue, slashDelimitedPath) => {
```
tilleryd: I think we should push back on the gm-fabric-go dudes and ask for better JSON keys and as such just ask to standardize on the keys we expect to be most useful in the dashboard.

https://github.com/DecipherNow/gm-fabric-go/issues/134

src/utils.js
```diff
@@ -193,8 +206,117 @@ export const getThreadCounts = createSelector(getCurrentThreads, (threadsTable =
...
+  });  
+  console.log(normalizedTimeseriesData);
+  return normalizedTimeseriesData;
+}
+
+/**
+ * Utility function that takes the raw metrics data scraped from the AWS endpoints, normalizes the data in the
+ * form expected by the React UI components, and restructures as timeseries data using the UNIX timestamp of when the
+ * endpoints were polled.
+ * @param {*} rawScrapedMetrics
```
tilleryd: Shouldn't `*` be `Object`?

src/utils.js
spmcbride1201: Something like this? https://github.com/facebookincubator/create-react-app/issues/578#issuecomment-277843310

PR #133, opened 2017-07-20
Relative Path Bug Fix

### LINKED ISSUE #112: Use of Relative paths for loading assets causes hard loads to fail on deep routes in the client-side router.
If I load `https://private.deciphernow.com/services/ess/1.0/gmadmin/dashboard/jvm`, then the app attempts to load the CSS bundle from `https://private.deciphernow.com/services/ess/1.0/gmadmin/dashboard/static/css/main.***.css`. This causes me to suspect that the app is currently using `./` as the baseURL for the client router, when it should really be the URL set in the `index.html` head

image

spmcbride1201:
Prior to (the) TheNextWeb conference, I had to hack together a solution to be able to host this app in a deeply-nested path. The end result worked, but had issues with some deeply nested client side routes. This PR represents my attempt to put together a better solution. It involves changing how I injected string templates into index.html and the JS bundle, changes to the BASH script that replaces the template, and changes to the client-side logic for inferring scrape target endpoints based on where the app is hosted. The README has been modified to adjust these changes.

In the course of this work, I encountered unusual `node-gyp` errors (actually related to an NPM 5 bug, but I didn't know it at the time). In the course of "fixing" these errors, I blew away my NODE_MODULES directory and the package lock file, reran `npm install`, and went down a rabbit hold that involved the following:
1. Removing react-a11y based on the inclusion of an alternate solution in CRA
2. Removing eslint middlewares that were redundant with CRA
3. Removing other unused modules (and then reinstalling one because I was overeager)
4. Updating CRA to 1.0.10 to pick up relevant bug fixes.

As this is a hot fix, I've tentatively deployed the dashboard, so you can see this code running at:
https://private.deciphernow.com/services/ess/1.0/gmadmin/summary

Good Times!
Resolves #112


package.json
```diff
@@ -20,38 +19,34 @@
...
   },
   "devDependencies": {
     "enzyme": "~2.8.2",
     "eslint-config-prettier": "^2.3.0",
-    "eslint-plugin-flowtype": "^2.35.0",
-    "eslint-plugin-import": "^2.7.0",
-    "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-prettier": "^2.1.2",
     "eslint-plugin-react": "^7.1.0",
+    "jest-cli": "^20.0.4",
```
spmcbride1201: totally mysterious why I needed to install this now, but CircleCI wouldn't run my tests otherwise.

src/utils.js
```diff
@@ -238,9 +239,9 @@ export function getRuntime() {
 export function generateEndpoints() {
   switch (getRuntime()) {
     case "JVM":
-      return ["admin/metrics.json"];
+      return [getBasename().replace("/gmadmin/", "/admin/metrics.json")];
```
spmcbride1201: I sorta explain the logic of these functions in the README. In the future, I plan to provide a mechanism that these can be set manually / adjusted via the settings component in the app.

src/registerServiceWorker.js
```diff
@@ -8,40 +8,97 @@
 // To learn more about the benefits of this model, read https://goo.gl/KwvDNy.
```
spmcbride1201: All of these changes are just copypasta following the CRA upgrade guide

.eslintrc
```diff
@@ -27,7 +27,9 @@
     "react/self-closing-comp": 1,
     "react/sort-comp": 1,
     "react/sort-prop-types": 1,
-    "react/wrap-multilines": 0
+    "react/wrap-multilines": 0,
+    "jsx-a11y/href-no-hash": "off",
+    "jsx-a11y/anchor-is-valid": ["warn", { "aspects": ["invalidHref"] }]
```
spmcbride1201: I had to add these to silence ESLINT errors. There's an active GitHub Issue open, so I view this as a temporary workaround to hide meaningless linting errors.

src/utils.js
```diff
@@ -254,7 +255,7 @@ export function generateEndpoints() {
 export function generateThreadsEndpoint() {
   switch (getRuntime()) {
     case "JVM":
-      return "admin/threads";
+      return getBasename().replace("/gmadmin/", "/admin/threads");
```
shindigira: Verified that the proper endpoints are being set.   

screen shot 2017-07-21 at 4 43 32 pm

src/style/index.less
```diff
@@ -16,8 +16,12 @@
...
 @import 'components/reactInputRange.less';
 @import 'components/recharts.less';
-@import '../../node_modules/react-grid-layout/css/styles.css';
-@import '../../node_modules/react-resizable/css/styles.css';
+
+// Hard Copied Stylesheets from 3rd party react components
+// These styles might become stale or out of sync with the
+// components that they reference
+@import 'components/react-grid-layout.css';
+@import 'components/react-resizable.css';
```
shindigira: This was giving me an issue when I `npm install`ed the latest modules but this is a good workaround.

PR #148, opened 2017-07-24
feat: Add basic support for rendering Lyft Envoy Metrics.

spmcbride1201:
To test this, first install and run the Envy "Front Proxy' environment

```bash
git clone git@github.com:lyft/envoy
cd envoy/examples/front-proxy
docker-compose up --build -d
```

Confirm that you see metrics at localhost:8001/stats

Then run `npm start` in `gm-fabric-dashboard`

You should only see Explorer and Setting appear and Explorer should have metrics populated from Envoy.


src/store.js
```diff
@@ -28,19 +33,43 @@ import defaultDashboards from "./json/dashboards.json";
...
-    })
-    .then(json => Actions.fetchMetricsSuccess(json))
-    .catch(err => Actions.fetchMetricsFailure(err));
+  const runtime = getRuntime();
+  // This block allows us to direclty poll Envoy metrics from a single source
+  // Everything beyond the first endpoint is discarded
+  // The data is transformed from statsd format into a flat object of key/value pairs
+  if (runtime === "ENVOY") {
+    axios
+      .get(endpoints[0], { responseType: "text" })
```
dgoldstein1: what is `endpoints[0]`?

src/store.js
```diff
@@ -28,19 +33,43 @@ import defaultDashboards from "./json/dashboards.json";
...
+      .then(jsons => {
+        let results = {};
+        jsons.forEach(json => {
+          results = { ...results, ...json };
+        });
+        return results;
+      })
+      .then(json => Actions.fetchMetricsSuccess(json))
+      .catch(err => Actions.fetchMetricsFailure(err));
+  }
```
dgoldstein1: These changes look good!~ very readable

src/store.js
```diff
@@ -111,15 +140,34 @@ Effect("initLocalForage", () => {
...
+          defaultDashboards,
+          runtime
+        );
+        console.log("DBs: ", dashboardsForRuntime);
+        if (Object.keys(dashboardsForRuntime).length > 0) {
+          Actions.updateDashboardsRedux(dashboardsForRuntime);
+        }
       }
     })
     .catch(err => console.log("fetchDashboards failed with ", err));
```
dgoldstein1: did you want to keep the loging statements in here?

src/store.js
spmcbride1201: endpoints is as array of strings containing different addressable endpoints. Previously, I handled multiple endpoints in the same promise chain to guarantee that they were associated with an identical timestamp, but I've had to back off that because of key collisions. I added a few more comments to explain this.

src/store.js
spmcbride1201:
```js
      .then(statsdTextFile => {
        let results = {};
        statsdTextFile
          .split("\n")
          .map(kv => kv.split(": "))
          .forEach(([key, value]) => {
            if (key) {
              results[key] = Number(value);
            }
          });
        return results;
```

I cleaned up the previous step before this a bit more. Looking at it again, kv[0] and kv[1] wasn't easy to understand.

src/store.js
spmcbride1201: I'd like to keep them in for a while longer. I find them helpful to debug unexpected state issues.

PR #151, opened 2017-07-25
Issue #141: Add zipkin tracing into the environment

joshua-rutherford:
@spmcbride1201 adding tracing via zipkin was fairly straight forward.


joshua-rutherford:
And Zipkin is exposed on `localhost:9411`.

joshua-rutherford:
@cstrahan and @alecholmez, figure I should keep you abreast of this too.  So I've provided Sean with a very simple setup here.  He's got a dead simple Ruby service that is sitting behind an Envoy sidecar (not actually in the same container as it's just easier for our purposes) and that is sitting behind an edge Envoy proxy.  Lastly this pull request adds tracing via a Zipkin server.  This bit is open for discussion but it was easy with Envoy so I through it in.

I am able to get @alecholmez example service compiled and running and could have used it here, but figured having a simple environment that doesn't require much tooling would be best.

PR #155, opened 2017-07-26
Toolbar Right

### LINKED ISSUE #154: Implement Navbar Right
Implement navbar component to contain breadcrumbs, link to settings, and potential future components like notifications and accounts.

shanberg:
- Adding app-toolbar component
- Refactoring navbar into app-brandbar

Closes #154


src/components/AppBrandBar.js
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <nav className="app-header app-brandbar">
+      <div className="brand">
+        <img
+          alt="Decipher Logo"
+          className="brand-logo"
+          src={decipherLogo}
+          style={{ width: "20px" }}
+        />
+        <span className="brand-text" id="logo-text">{'Fabric'}</span>
```
spmcbride1201: Please get rid of the {' explicit stuff around your string. We do this implicitly everywhere else, and I think this makes things more readable / similar to HTML.

src/components/AppBrandBar.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import decipherLogo from "../images/decipherLogo.svg";
+import { NavLink } from "react-router-dom";
```
spmcbride1201: Unused import. Please remove.

src/components/AppBrandBar.js
```diff
@@ -0,0 +1,31 @@
...
+        <span className="brand-text" id="logo-text">{'Fabric'}</span>
+      </div>
+
+      <a
+        className="app-version"
+        href="https://github.com/DecipherNow/gm-fabric-dashboard/blob/master/CHANGELOG.md"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        {'0.1.1'}
```
spmcbride1201: ditto like above comment.

src/components/AppToolBar.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import decipherLogo from "../images/decipherLogo.svg";
+import { NavLink } from "react-router-dom";
+
+const AppToolBar = () => {
+  return (
+    <nav className="app-header app-toolbar">
+      <ol className="app-breadcrumbs">
+        <li className="app-breadcrumb"><a href="">{'Fabric'}</a></li>
```
spmcbride1201: Please replace the anchor tags with the react-router-dom `Link` element for these three list elements in the breadcrumb. You'll have to add this to the import on line 3. It has a required `to="some/path/"` prop that functionally the same as `href` on an anchor tag. You can just pass an empty string for now. My goal is just to prevent a hard refresh of the bunch when these are clicked.

src/components/AppToolBar.js
```diff
@@ -0,0 +1,25 @@
...
+        <li className="app-breadcrumb"><a href="">{'Service'}</a></li>
+        <li className="app-breadcrumb"><a href="">{'Instance'}</a></li>
+      </ol>
+
+      <NavLink
```
spmcbride1201: Not critical, but you should check out `activeClassName` as a means to style this link when the nav item matches the current route. https://reacttraining.com/react-router/web/api/NavLink/activeClassName-string

src/style/components/navbar.less
```diff
@@ -52,41 +54,98 @@
...
+      @background-color: @toolbar-background-color;
+      @degree: 6%;
+      @mode: darken
     );
-    .mixin-button-height(@navbar-height * .85);
-    .mixin-temporary-fix-for-button-icon-placement();
+    .mixin-button-height(@app-header-height * .85);
+  }
+
+  .uk-button + .uk-button {
```
spmcbride1201: image

I'm not sure if the settings buttons are getting affected here, but it looks strange that they aren't centered relative to the headers. What do we need to do to make this look cleaner?

src/components/AppToolBar.js
shanberg: This shouldn't be necessary for the breadcrumbs, as they should be generated by the routing itself, right?

src/style/components/navbar.less
shanberg: Settings should all be better in a coming PR.

spmcbride1201: Opened #158

PR #156, opened 2017-07-27
Functional Docker-Compose piping Envoy metrics to console, Graphite, or Prometheus

spmcbride1201:
## Test one: Pipe Envoy's statsd metrics to console

```bash
cd docker
docker-compose -f docker-compose-dummy.yml up
```
and then look for console log messages from DUMMY with statsd metrics
```bash
docker-compose -f docker-compose-dummy.yml down
```

## Test two: Graphite
```bash
docker-compose -f docker-compose-graphite.yml up
```
Then go [here](http://localhost:1337) and look for Envoy metrics.
You should see metrics change by hitting [this service](http://localhost:8080/services/clock/1.0/now)
```bash
docker-compose -f docker-compose-graphite.yml down
```
## Test three: Prometheus
```bash
docker-compose up
```
Then go [here](http://localhost:9091) and search for Envoy metrics.
You should see metrics change by hitting [this service](http://localhost:8080/services/clock/1.0/now)
```bash
docker-compose down
```


joshua-rutherford:
I assume in the above comment you meant the following in test two: 
```
docker-compose -f docker-compose-graphite.yml up
```

spmcbride1201:
Nice catch. Thank you for your attention to detail!

docker/docker-compose.yml
```diff
@@ -1,33 +1,47 @@
...
+      fabric:
+        ipv4_address: 172.16.238.08
     ports:
       - 7080:7080
       - 7081:7081
 
   edge:
 
     build: ./edge
+    volumes:
```
joshua-rutherford: For consistency we should either mount these as volumes on all of them or just build the image with them as I've done.  Volumes makes changes a bit easier to implement (i.e., no need to rebuild) so I vote for your approach I think.

docker/dummy/index.js
```diff
@@ -0,0 +1,18 @@
+const dgram = require("dgram");
```
joshua-rutherford: This docker setup is a polyglot's dream.  Ruby and node services like ebony and ivory.

docker/edge/Dockerfile
```diff
@@ -1,5 +1,3 @@
 FROM lyft/envoy:latest
 
-COPY etc/envoy/envoy.json /etc/envoy/envoy.json
-
-CMD ["envoy", "-c", "/etc/envoy/envoy.json", "--service-cluster", "edge"]
+CMD ["/usr/local/bin/envoy", "--config-path", "/etc/envoy.json", "--service-cluster", "edge", "--service-node", "0", "--log-level", "debug"]
```
joshua-rutherford: Do we have an understanding of how the `--service-cluster` and `--service-node` get reflected in the stats?  I don't see the `--service-node` bit in there but I know Envoy requires it to export the stats.

joshua-rutherford:
@spmcbride1201 the following comment is not against your pull request but general points we must understand or work as a team.

@cstrahan, @alecholmez and @dougford bringing you in on this too.

Well I'm happy I went to bed Tuesday night instead of trying to get this noodled out.  Following you debugging logic (and it's sound) I wouldn't have slept at all.  Here are my concerns at this point:

1. The static IP address requirement when mixing Envoy with the StatsD Exporter are a non-starter.  I just don't think we can accept that requirement.  Any timeline on the StatsD Exporter next release that will support TCP and thus DNS based resolution from Envoy?

2. I can see where the service is indicated in the metrics (i.e., we named the `stat_prefix` for the clock service `ingress_http`) however, I may define multiple routes in the `route_config` and that could mean that multiple services get the same prefix.  We need to understand this and determine the configurations required to avoid that confusion.

3. I cannot tell what endpoint is being hit at all.

spmcbride1201:
No idea about release date for stats_exporter 0.5.0, but here's the relevant PR in their project. https://github.com/prometheus/statsd_exporter/pull/71

spmcbride1201:
Currently, stats_explorer is using whatever their default transforms are. They strongly suggest writing a custom mapping definition. https://github.com/prometheus/statsd_exporter#metric-mapping-and-configuration

As Josh mentions above, I'm pretty sure that there are some tagging issues from Envoy as well. I believe I'm only pushing stats from the Edge, not the sidecar. I was a bit gun shy about label collisions. The Envoy documentation for this is pretty much non-existent, and there aren't any community blog posts out there, so it's a bit of a shot in the dark.

docker/docker-compose-dummy.yml
```diff
@@ -0,0 +1,60 @@
...
+
+networks:
+
+  fabric:
+    driver: bridge
+    ipam:
+      driver: default
+      config:
+      - subnet: 172.16.238.0/24
+      - subnet: 2001:3984:3989::/64
```
tilleryd: please indent the array items under `config`

tilleryd:
I ran through each test (with @spmcbride1201 guidance) and verified that all containers started up successfully.  Navigated to the Prometheus and Graphite UIs and viewed metrics/graph data for the clock service.

docker/docker-compose-dummy.yml
tilleryd: i'm a pedantic ass

PR #160, opened 2017-07-27
Implement nav sibling dropdown

shanberg:
Dummy main nav dropdown to contain siblings, in the current case meaning other instances.


src/components/SummaryBar.js
```diff
@@ -33,6 +34,20 @@ function SummaryBar({ dashboards, metrics, interval, runtime }) {
...
         </a>
+        <div
+          data-uk-dropdown="mode: click; pos: bottom-justify; boundary: .nav-widget; boundary-align: true; offset: 1;"
+          className="nav-siblings-dropdown"
+        >
+          <ol className="instances">
+            {Array(90).fill("shanberg").map((val, idx) =>
+              <li key={idx}>
+                <Link to="">
+                  Instance {Math.floor(Math.random() * 9)}
```
spmcbride1201: Now that I see this, I think random looks odd. If you change this to what I have below, it should increment from 1 to 90
Instance {idx + 1}

shanberg: Should be fixed now.

PR #168, opened 2017-07-27
docs: Updated README with instructions for different runtimes

spmcbride1201:


tilleryd:
Note related to this PR but this sentence needs improvement "Because new major versions of the Node.js runtime are releases major versions every six months …”.  Otherwise LGTM.

PR #170, opened 2017-07-27
Improve look of mocked out service and instance component

spmcbride1201:
Rather than hiding the service and instance component, this allows the service name to be configured via a index.html head attribute.


shanberg:
To clarify, my intention for the nav widget is for the first button to be the universal 'Back/Up' button, and the second to do double-duty as the page title and the navigation to page siblings. So it won't always be Service + Instance, if that makes a difference.

PR #174, opened 2017-07-28
Update dependencies and replace stylelint with prettier's CSS support

spmcbride1201:


src/style/components/routeBar.less
```diff
@@ -3,8 +3,8 @@
 @import '../mixins.less';
 
 // Local Variables
-@route-bar-background: #FFF;
-@route-bar-card-border: #CCC;
+@route-bzar-background: #fff;
```
shanberg: The typo here prevents my CSS from building.

As an aside, why did all tests pass when the CSS wouldn't build?

spmcbride1201: Fixed. I'm going to have to file an issue for making CI fail when preprocessing doesn't work as expected.

PR #179, opened 2017-08-01
Merging initial sass fixes upstream

spmcbride1201:


package.json
```diff
@@ -96,9 +94,18 @@
...
+      "prettier --write",
+      "git add"
+    ],
+    "*.scss": [
+      "prettier --write",
+      "git add"
+    ],
+    "*.less": [
+      "prettier --write",
+      "git add"
```
shanberg: We're still moving theoretical `less` files around even though there won't be any. Not a real issue, just a note.

src/style/components/nav-widget.scss
```diff
@@ -48,13 +48,13 @@ $go-up-icon-size: $padding-base * 2;
...
   .nav-go-up,
   .nav-siblings {
-    @include mixin-button-style(
-      $nav-widget-text-color,
-      $nav-widget-background-color,
-      $nav-widget-background-color,
-      lighten
-    );
-    @include mixin-button-height($padding-base * 4);
+    // @include mixin-button-style(
```
shanberg: I've commented out all the places I'm loading the button style builder. The button styles are in progress and will be fixed in a later PR.

src/style/components/routeBar.scss
```diff
@@ -26,7 +26,7 @@ $active-route-bar-card-background: $decipher-green;
 
   // Conflicts with UK-Card styles.
   .route-bar-card-body {
-    @include uk-card-body();
+    // @include uk-card-body();
```
shanberg: I will need to fix the cards and other `less`-dependent imports from UIkit. Should not be a significant issue, just a separate PR.

spmcbride1201: We're pulling in the UIKit v3 scss source correctly, right? Is any of that stuff misconfigured?

PR #181, opened 2017-08-01
Shifted AJAX and Local Storage stuff into Web Workers

spmcbride1201:
Also, deleted the default service worker provided by CRA.


src/index.js
```diff
@@ -6,10 +6,16 @@ import { ConnectedRouter as Router } from "react-router-redux";
...
 import Icons from "uikit/dist/js/uikit-icons";
 
-import registerServiceWorker from "./registerServiceWorker";
 import Container from "./components/Container";
 import store from "./store";
 import { getBasename } from "./utils";
+import PromiseWorker from "promise-worker";
+
+// Create a Web Worker and explicitly set reference as global variable
+const AjaxWorker = require("worker-loader!./workers/ajax.js");
```
tilleryd: Any particular reason to use the CommonJS style of module import instead of ES6?  If Webpack is converting to UMD then perhaps it's best to just stay consistent with one style.

src/workers/ajax.js
```diff
@@ -0,0 +1,54 @@
+var registerPromiseWorker = require("promise-worker/register");
```
tilleryd: Yeah more mixing of `require` and `import`.  Is this necessary?

src/workers/localStorage.js
```diff
@@ -0,0 +1,52 @@
...
+        name: `grey-matter-fabric-${basename}`,
+        description:
+          "Persistent storage of Grey Matter Fabric dashboards and settings"
+      });
+      break;
+    case "getDashboards":
+      const { runtime } = message;
+      return localforage
+        .getItem("dashboards")
+        .then(savedDashboards => {
```
tilleryd: I like this async feature of `localForage`.

tilleryd:
@spmcbride1201 Review complete.  I confirmed that JumpState actions are being triggered by the WebWorker accordingly.  Aside from a couple small comments the code looks good.

PR #182, opened 2017-08-01
fix: Restore eslint.

spmcbride1201:


tilleryd:
@spmcbride1201 This config seems to be working as intended.  When I start the app it fails to compile due to `console.log` statements in `store.js`.  I like that the app doesn't even start if there are lint errors.  It forces engineers to fix the issues.

PR #184, opened 2017-08-02
New buttons, fixed Sass

shanberg:


src/components/PollingSettings.js
```diff
@@ -21,8 +21,8 @@ class PollingSettings extends Component {
   render() {
     const { isPolling } = this.props;
     const buttonClass = isPolling
-      ? "uk-button uk-button-vertical uk-button-danger"
-      : "uk-button uk-button-vertical";
+      ? "btn btn-type-primary btn-icon-size-xl btn-outline-none btn-orientation-vertical"
+      : "btn btn-type-primary btn-icon-size-xl btn-outline-none btn-orientation-vertical";
```
spmcbride1201: I'm going to DRY this code up for you a bit. I'd ask you to do it, but I know you're enroute. Please be sure to stop working in this branch and rebrach off development.

src/components/SettingsGrid.js
```diff
@@ -35,7 +32,7 @@ function SettingsGrid({ settings }) {
             </div>
 
             <button
-              className="uk-button"
+              className="btn"
```
spmcbride1201: We might want to make this something like gm-button down the road if this becomes something we share across projects.

src/style/components/buttons.scss
```diff
@@ -0,0 +1,238 @@
+// components/buttons.scss
```
spmcbride1201: We'll definitely need to document this stuff.

PR #188, opened 2017-08-03
feat: Reimplement GMLineChart and GMGrid using Dygraphs

spmcbride1201:
This commit involved modifying components and utility functions required to render grids and line charts with the dycharts library.


src/utils.js
```diff
@@ -1,467 +0,0 @@
...
-            threadItem =>
-              threadItem.state === "TERMINATED" ||
-              threadItem.state === "BLOCKED" ||
-              threadItem.state === "NEW"
-          ).length
-        : 0,
-      all: threadsTable.length
-    };
-  }
-);
```
jmcudd: Nice!

src/utils/threads.js
```diff
@@ -0,0 +1,68 @@
...
+            threadItem =>
+              threadItem.state === "TERMINATED" ||
+              threadItem.state === "BLOCKED" ||
+              threadItem.state === "NEW"
+          ).length
+        : 0,
+      all: threadsTable.length
+    };
+  }
+);
```
jmcudd: What are threads used for here? Is this for managing your workers you implemented earlier?

src/components/DygraphContainer.js
```diff
@@ -0,0 +1,67 @@
...
+  static propTypes = {
+    timeSeries: PropTypes.array.isRequired
+  };
+
+  componentDidMount() {
+    const [data, options] = this.props.timeSeries;
+    this.drawChart(this.div, data, options);
+  }
+
+  componentWillReceiveProps(nextProps) {
```
jmcudd: This is nice way of updating charts with new data! I might be borrowing a lot from this when I eventually refactor our chart.js in my project.

PR #189, opened 2017-08-04
feat: Deploy cleanup

spmcbride1201:
Removing comments, falling back to hash history, fixing bug in getDyngraphOfValue, setting summary cards to open by default, and adding error handling for fetchThreads


spmcbride1201:
I like it quite a bit. It fixes all of my nibbles with vanilla fetch. 👍

PR #206, opened 2017-08-07
New Fonts

shanberg:
New fonts from Google Fonts, which are close-ish to Decipher branding and may improve usability.


public/index.html
```diff
@@ -22,7 +22,7 @@
       homescreen on Android. See https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/
     -->
   <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
-
+  <link href="https://fonts.googleapis.com/css?family=Nunito:400,600|Rubik:400,500" rel="stylesheet">
```
spmcbride1201: This is fine as a temporary fix, but we're going to want to locally embed all static assets into the project. I opened #209 for this and plan to do this myself later.

PR #211, opened 2017-08-07
feat: Implemented configurable back button.

spmcbride1201:
Added meta tag containing a URL, a helper function for retrieving the value, and an achor tag that uses it.


spmcbride1201:
@shanberg - It's configurable at deployment. I don't want to hard code it in a build if it only applies to our environment. Try changing `__BACK_BUTTON_URL__` to 'https://private.deciphernow.com/services/'

shanberg:
Sounds good then. LGTM.

PR #216, opened 2017-08-08
Implement new routes component

spmcbride1201:
Also build out mock-jvm infrastructure.

In separate shell sessions, please run `npm run mock-jvm` and `npm start`.


src/components/RoutesTableLineItem.js
```diff
@@ -0,0 +1,94 @@
...
+    this.setState({ isOpen: !this.state.isOpen });
+  };
+
+  render() {
+    return (
+      <li
+        className={"selectable open-" + this.state.isOpen}
+        onClick={this.toggleDrawer}
+        onKeyDown={evt => {
+          if (evt.keyCode === 13) {
```
dgoldstein1: I wasn't able to toggle drawers by typing enter, but this is a small thing

src/components/RoutesTableLineItem.js
```diff
@@ -0,0 +1,94 @@
...
+        <div className="routes-table-route">
+          {this.props.route}
+          <div className={"route-viz-bar err-pc-" + this.props.errorRate}>
+            <div className="route-viz-fill" style={{ width: "50%" }} />
+          </div>
+        </div>
+        <div className="routes-table-sparkline">
+          <Sparklines
+            data={this.props.requestsPerSecond_sparkline}
+            height={32}
```
dgoldstein1: set `32` in line 51 to variable (avoiding magic numbers)

src/components/RoutesTableLineItem.js
```diff
@@ -0,0 +1,94 @@
...
+              style={{
+                stroke: "currentColor",
+                strokeWidth: 1,
+                fill: "none"
+              }}
+            />
+            <SparklinesReferenceLine
+              style={{
+                stroke: "grey",
+                opacity: "0.4"
```
dgoldstein1: maybe make `sparkLinesReferenceLine` a css class? (avoiding hardcoded values)

src/style/components/forms.scss
```diff
@@ -0,0 +1,48 @@
...
+    border: 0;
+  }
+
+  &:focus {
+    outline: none;
+    box-shadow: inset 0 0 0 rgba(255, 255, 255, .5),
+      0 0 0 1px transparentize($color-highlight, .25),
+      0 0 0 4px transparentize($color-highlight, .75);
+  }
+}
```
dgoldstein1: The app is looking awesome!!!

src/style/components/routesTable.scss
```diff
@@ -0,0 +1,102 @@
...
+  .div-table-body {
+    .routes-table-sparkline {
+      height: 32px;
+      padding-top: 0;
+
+      svg {
+        height: 100%;
+      }
+    }
+  }
```
dgoldstein1: As a COMPLETELY personal opinion, I thought the lines looked a little thin in the routes table and the sidebar

src/utils.test.js
```diff
@@ -3,8 +3,8 @@ import state from "./json/mockReduxState";
 const { getVisibleThreads, getThreadCounts } = require.requireActual(
```
dgoldstein1: is there a reason these aren't es6 imports?

src/utils/routes.js
```diff
@@ -0,0 +1,104 @@
...
+      _.set(routeList, [routePath, httpVerb], [...previousValue, route]);
+      // Otherwise add the value as the sold value of an array at that key
+    } else {
+      _.set(routeList, [routePath, httpVerb], [route]);
+    }
+  });
+  return routeList;
+}
+
+export const getRoutesTable = createSelector(
```
dgoldstein1: could you add a comment what this does? unclear on first look

src/utils/routes.js
```diff
@@ -0,0 +1,104 @@
...
+            ),
+            errorRate
+          }
+        );
+        routes.push(localObj);
+      });
+    });
+    return routes;
+  }
+);
```
dgoldstein1: perhaps separate lines 64-101 into separate helper function then do `Object.keys(routesTree[routePath]).forEach(newFunction)`? just a thought to reduce the size of this

src/utils/sparklines.js
```diff
@@ -23,12 +23,21 @@ export function getSparkLineOfValue(metrics, key) {
  */
```
dgoldstein1: can you specify what `Array` means in the `@return` jsdoc header (unclear what `[0,0]` would be-- I'm assuming index = 0 is old, and index = 1 is new?

src/utils/sparklines.js
```diff
@@ -23,12 +23,21 @@ export function getSparkLineOfValue(metrics, key) {
...
+      const lastTime = timestamps[idx - 1];
+      return idx === 0
+        ? 0
+        : Math.round(
+            (Number(value) - Number(last)) / ((currentTime - lastTime) / 1000)
+          );
     })
     .slice(1);
+  return results;
 }
```
dgoldstein1: this code looks great 👍 but definitely needs testing

src/components/RoutesTableLineItem.js
spmcbride1201: Good catch. I'm pretty sure the way focus is handled is thoroughly broken. I'm going to open an issue for this.

src/components/RoutesTableLineItem.js
spmcbride1201: I feel like this is a bit overkill, especially since I'd have to declare the intermediate variable outside of the return block. You don't think that the prop name give enough semantic meaning here?

src/style/components/routesTable.scss
spmcbride1201: @shanberg FYI

shanberg: @dgoldstein1 Which lines? Lines in the graphs? Lines between table items?

src/components/RoutesTableLineItem.js
spmcbride1201: This actually sent me down a bit of a rabbit hole, and I wanted to report back to share the lessons learned. Because we're using a 3rd party component, we have to pass our styles as props here so that the component is able to apply the styles properly to its children components. If we add a className, it will get applied to the outermost DIV, not the proper internal piece of the component.

dgoldstein1: good to know :) that probably just saved me 30min - 1hr

src/style/components/routesTable.scss
dgoldstein1: 1. the lines between items in the side bar (I.e. line between routes)
2. lines between the table items

src/utils.test.js
spmcbride1201: Just following the Jest docs. I wouldn't know how to call `requireActual` with ES6

src/components/RoutesTableLineItem.js
dgoldstein1: Sounds good, no right answer here just my immediate thought when I saw a hard coded int

PR #224, opened 2017-08-09
fix: Local storage now cleanly initializes

spmcbride1201:


tilleryd:
LGTM.  On page load with an empty IndexedDB, the appropriate data is set in localStorage to display all dashboards.  Code looks good too.

PR #252, opened 2017-08-10
fix: Unescape slashes pulled from the URL

### LINKED ISSUE #191: Breadcrumps needs to "unescape" URLS
Currently, the breadcrumbs displays '%2F'

image

spmcbride1201:
Small change for a bug that we can't currently hit, but we might again if we inject paths as route params when using pushstate.

Resolves #191


shanberg:
How do you recommend I test this?

spmcbride1201:
Sorta impossible right now. Just make sure the breadcrumbs aren't broken and take it on faith essentially.

PR #271, opened 2017-08-11
feat: Release 0.3.0

spmcbride1201:
This is already deployed to https://private.deciphernow.com/services/ess/1.0/gmadmin/#/summary


joshua-rutherford:
I'm not sure how you want me to capture my comments with this as some may be existing issues or are just plain dummy data for demo but here they are:

1) On the "Summay" tab for every service, the "Host" and "Port" are pretty much just wrong.  That is the host and port of the edge (actually maybe not even that as it's port 80 and I'm SSL).  Not sure if you can correct that now, but it make it look like the service instance is running on the edge on HTTP). 

2. On the "Summary" tab if we have host name, we should have IP Address as well (DNS sometimes breaks or doesn't exist at all).

3. Instance selection dropdown has no effect.

4. Personally, the left navigation bar tries to provide too much information which makes understanding navigation difficult.  I'd prefer text only or at a minimum they should be collapsed by default.

shanberg:
@joshua-rutherford 
re: 4) Thanks for the UX feedback, that's extremely helpful to hear. Collapsed by default sounds reasonable, but I'm interested in what you think of the actual goal, execution aside, of making the left nav be a glanceable overview of the performance. All text is simple to do, but reduces context for the information, and some values may change rapidly. Do you think it could be a matter of choosing the correct values and data visualizations, or is it just not a helpful goal?

joshua-rutherford:
Personally, performance has so many aspects that I don't know how you choose what is the overarching thing that is important enough for that.  It may be useful to leave that to the user (i.e., have a "pin to navbar" checkbox for all graphs in a view with the default of none.  I personally would never pin any but I can't say no one would.

joshua-rutherford:
For that matter, you could have a "pin to summary" feature too.

joshua-rutherford:
And while your at it, can it be amphibious?

spmcbride1201:
@joshua-rutherford - Couple of notes about dummy versus live metrics.

The Host and Port are basically scraped from the browser, so it's the edge. However, my logic for inferring the port if it's not explicit is crappy. I basically fallback to 80 in all cases. To my knowledge, there's nothing in `metrics.json` that actually provides any of this info. I've presumed that I'd replace this with metrics from Zookeeper or Envoy.

Instance selection and the functionality of the ESS button going back to the services gateway are basically also semi-functional mocks. This working properly also hinges on getting better integration with Zookeeper or Envoy. 

We'll have to discuss what this will look like for initial delivery to a client, but if we're sticking with Zookeeper and NGINX as is, I suggest having NGINX template out to a JSON instead of HTML and pulling the service gateway page functionality into this single page app.

spmcbride1201:
The pinning feature seems like a good approach to me. I think that gets wrapped up in the whole lifecycle discussion of how we persist these custom dashboards.

shanberg:
@joshua-rutherford I'll see what we can do. Optimizing for high external pressures will reduce performance in low-pressure environments, so we may need to rethink the exoskeleton once we begin space-based operations, but that's not in the roadmap for another few major versions.

For demo purposes, I recommend we keep some extras in the sidebar, and let the items be collapsed by default.

I'll post an issue for discussion of default items, and another for general item pinning.

joshua-rutherford:
https://www.youtube.com/watch?v=O4RLOo6bchU

PR #281, opened 2017-08-15
WIP: First crack at creating text color styles from value of expressed data

shanberg:
New `.exp-text` class on an element, combined with `data-value-ratio` of 0–100 should change the text color to a point on a gradient corresponding to the value ratio. By default, 0 should be red and 100 should be green, but we can create a gradient of any number of different colors.

Next step here is to build the features into components where they're useful, such as ReadoutItem.

Example usage:  `<span className="readout-feature-value exp-text"  
data-value-ratio="{valueAsPercentOf100}">  
{value}  
</span>`

screen shot 2017-08-15 at 4 45 41 pm

screen shot 2017-08-15 at 4 45 25 pm


spmcbride1201:
Sorry to bust your bubble first thing on a Monday, but I suspect that passing data attributes to CSS isn't going to be flexible enough for our purposes. Let's circle back with @tilleryd to discuss.

Impressions of Current PR: 
- For a lot of these attributes (or perhaps all), the relationship between safe, warning, and error type states aren't going to exist on a linear continuum. I would expect breakpoints between discrete states (either in percentage or absolute values). At the start, these would be configurable by the user. Later, once we start implementing integrations with things like GMSense, we quite possibly will be opening up an API that allows GM Sense to dictate these breakpoints based on trend analysis. Thinking about this, this suggests that the CSS should offer more traditional `nominal`, `warning, `error` type selectors (perhaps with appropriate animations between those states if we retain your gradient-approach)
- I think the compiled code for these sorts of functions really embiggens our CSS. Something might have misfired (I only see `data-value-ratio` up to 24, not 100), but assuming that every possible value needs to be precompiled as CSS, we're talking about 101 selectors here. It seems much more sane to me to move some of these CSS selectors into JS-land to DRY this up. E.g., I'd keep `.exp-text`, drop the data-attribute, and add a utility function in JS-land to manipulate `background-position-y` as an in-line style. That would take us down to one selector, and allow us to be more flexible with how we position these gradients (based on breakpoints, non-linear distributions, and all that kinda stuff). 

I have this gut feeling that if we want to really be dynamic and granular with our styles, we're going to have to take a look at some sort of css-in-JS solution, but that's essentially a sidebar discussion at this point.

shanberg:
I don’t mind at all, I’m all for the better and leaner solution, and this PR is only what I came up with given my skillset and priorities. It’s a stopgap solution.

spmcbride1201:
After discussing this issue with @shanberg offline, we decided to close this PR and shift his initial work on the Expressive Text functionality to a gist, which is at https://gist.github.com/spmcbride1201/8cd67d5c998e123d5ac7062536fe771c. We'll revisit this in the future if needed.

PR #285, opened 2017-08-15
feat: Format large numbers

### LINKED ISSUE #201: Add commas to large numbers

spmcbride1201:
Using toLocaleString without options allows us to inherit the OS-level locale settings. This seems like a pretty good approach.

Resolves #201


shanberg:
I think this will need to be followed up by some UX consideration for what numbers should be truncated (and when and to what degree), but this is a solid step forward.

PR #306, opened 2017-08-17
Color Refactor

shanberg:
1. **Instance Summary bar dashboard row now has more color.**

2. **UI Colors are now based on intended contrast value.**
What used to be `color: #000` should now be `contrast-color($background-color)`, which will output to either `color: #fff` or `color: #000` depending on what `$background-color` is. This makes adjusting element styles as simple as changing a single value, and its contents should react appropriately.

3. **New color functions**
- `custom-color-range($start-color, $size, $variance)` will create a list of different colors of a given size, based on the start color.
- `darker-color($color, $percent)` and `lighten-color($color, $percent)` provide an intelligent means of darkening and lightening colors with appropriate modifications to saturation.
- `edge-color($color, $contrast)` provides a quick and consistent way to create a border color based on the background color.
- `contrast-color($background-color, $contrast)` provides a legible high-contrast color based on the background color of the element.

4. **We can now modify whole application color by modifying the** `$color-sidebar-background` **and** `$color-content-background` **variables.**


src/style/functions.scss
```diff
@@ -0,0 +1,77 @@
...
+
+// Derive a readable content color from the background color of an element
+
+@function contrast-color(
+  $background-color,
+  $contrast: 100%) {
+  @if lightness($background-color) > 50 {
+    @return mix(#000, $background-color, $contrast);
+  } @else {
+    // Increase the contrast base for inverted colors, because light is strange
```
spmcbride1201: This comment feels downright philosophical. :)

src/style/functions.scss
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+
+// Derive a subtle keyline color for element separators
+@function edge-color(
+  $background-color,
+  $contrast: 8%) {
+  // If the element is very dark, use a light edge color
+  @if lightness($background-color) < 10 {
+    $enhanced-contrast: $contrast*$dark-on-light-contrast-enhancement-ratio;
```
spmcbride1201: Can you please put some white space between the variables and the sass-script operators (* sign)? It's a bit harder to read without it. There are a few of them in these functions.

src/style/layout.scss
```diff
@@ -40,6 +40,17 @@ $sidebar-width-base: 230px;
   background-size: 300% auto;
   background-repeat: no-repeat;
   background-position: bottom center;
+  position: relative;
+
+  &:after {
+    content: '';
```
spmcbride1201: I'm a bit confused by what's going on in here. Is the border-right the significant part here? I'm sure there's something I'm missing, but perhaps a comment here would be good to explain to numb-skulls like me what this is doing.

src/style/variables.scss
```diff
@@ -20,15 +14,34 @@ $color-white: #fff;
...
+$dark-on-light-contrast-enhancement-ratio: 2;
+
+$color-content: transparentize(contrast-color($color-content-background), .1);
+$color-content-muted: transparentize($color-content, .2);
+$color-highlight: $color-brand-primary;
+
+// Color Ranges
+$colors-datarange-blue-to-red: #3a76b1, #6993c2, #95b2d2, #c4d2e1, #f1f1f1,
+  #e8cdd2, #dfa9b0, #d48491, #ca6070;
+$colors-datarange-blueberry-kiwi: #1d334c, #486488, #d4e3ef, #76a834, #d2e3b4;
```
spmcbride1201: I really dig these names.

src/style/typography.scss
```diff
@@ -53,7 +54,7 @@ small {
 }
 
 .uk-text-primary {
-  color: $decipher-green !important;
+  color: $color-brand-primary !important;
```
spmcbride1201: Kinda out of scope, but seeing this is making me start to wonder about extending UIKit? Aren't we just importing the UIKit root scss file somewhere, which allows us to @extend their stuff?

src/style/components/data-displays.scss
```diff
@@ -22,6 +27,19 @@ $readout-feature-primary-text-color-base: $color-white;
 
   > .readout {
     margin: $padding-base / 2;
+    flex-grow: 1;
+  }
+
+  @for $i from 1 through length($readout-dashboard-colors) {
```
spmcbride1201: This seems like a good candidate for reuse as a @mixin that perhaps takes the classname targeted as an argument, perhaps with a second argument to specify the number of children (which is then passed into a call into one of your nifty functions). It would be nice to make your cool color range stuff a bit easier to use, especially as you roll it out to other front-end devs that might have varied levels of scss knowledge.

shanberg: At the moment I'm using the dashboard as a proof-of-concept for this technique. I do foresee abstracting the method further.

src/style/functions.scss
spmcbride1201: I'm actually a bit surprised that Prettier seems to not be doing anything in SCSS as expected. I vaguely seem to recall Prettier doing stuff to SCSS at one point, but it seems broken.

shanberg: The `prettier-atom` does seem to work for me. I'm not always happy with the result, but I've gotten used to what it produces.

In some cases the whitespace between variables and operators is meaningful, so I understand Prettier not touching it.

PR #325, opened 2017-08-21
Componentize buttons

shanberg:


shanberg:
Excellent. I stabbed at that for a while and then decided this was the best I could do. I'm excited to see things like this take shape.

tilleryd:
I just couldn't get a HOC pattern to work here.  I'm not spending anymore time on it.  @shanberg feel free to merge this as is.

PR #326, opened 2017-08-21
Fix type and alignment properties of Readout component

shanberg:


tilleryd:
The only issue I see is in `SummaryGrid.js`.

```
<Readout type={"readout-primary"}>
  ...
</Readout>
```

This would lead to the class of `readout-readout-primary`.

shanberg:
@tilleryd Fixed

tilleryd:
LGTM

PR #330, opened 2017-08-21
Add optional height property to set height of GMLineChart

shanberg:
This allows us to adjust the height responsively with CSS rather than having to use a wrapper div with a height value set in the HTML.


tilleryd:
@shanberg Was it intentional to divide the height of the chart on the summary page by 2?  It was 250px and now it's 125px.

tilleryd:
@shanberg I think the margin around the charts needs attention.  If we do go with an increased left margin then a matching right margin would help balance it out.

screen shot 2017-08-22 at 11 18 08 am

shanberg:
@tilleryd Space to the right of the chart is now equal to the space allocated to the y-axis labels, so charts should generally appear centered now. More improvements can be made here, but better done in later polish passes.

Issue #318, opened 2017-08-18
Develop Mock Service Discovery Service API


spmcbride1201:
Use Node.js and `json-server` to mock out a dummy API for the microservice that @joshua-rutherford will then develop. This will help document the desired behavior of the API and allow the UI implement the interface in parallel while the actual service is developed.

spmcbride1201:
The mocked service isn't yet at full parity with the API, but it shall be enhanced as I require the unimplemented APIs.

PR #351, opened 2017-08-23
feat: Completed mock of interim SDS

spmcbride1201:
Implemented updated to interim SDS and all remaining routes and optional params.

Resolves: #318


tilleryd:
LGTM

PR #352, opened 2017-08-23
fix: Debounce search to stop polluting history API on keyup

spmcbride1201:
Previously, typing in the searchbar fired the search function on every key event, which filtered the search results and updated the URL. This PR debounces the search function using 1000ms to ensure that the history API doesn't get polluted with extraneous stuff.

Resolves: #314


shanberg:
@spmcbride1201 Would it be feasible to skip the delay if user presses `return`?

spmcbride1201:
It should be. Let me take a look.

PR #359, opened 2017-08-24
Better Colors

shanberg:
Service summary grid and line graphs both get new color schemes.


src/style/functions.scss
```diff
@@ -68,7 +68,8 @@
   $type: qualitative,
   $direction: to-dark,
   $size: 12,
-  $variance: .5
+  $variance: .5,
+  $base-color-2: null
```
spmcbride1201: I see that this is passed into mix and append. What do these functions do when this is passed in as null?

shanberg: Base color 2 is only used in the divergence mode of the function. I suppose in the future, the function should test to make sure the proper variables are set for each mode, or the function should be refactored into three separate functions.

Base color 2 is ignored in qualitative and quantitative modes, and a color value is required for divergent mode.

spmcbride1201: Ok, can you add a comment that says something along the lines of "$base-color-2 is a required attribute when $type is divergent, but it otherwise ignored"?

PR #360, opened 2017-08-24
Crisper active summary bar card outline

shanberg:
Remove the dimmer borders on the top and bottom of active card items to improve the contrast between the active item background and the container background.


shanberg:
`~ A R T ~ D I R E C T I O N ~ `

PR #362, opened 2017-08-24
Refactor padding styles

shanberg:
Paddings and margins are now determined more clearly on a scale. This should make it easier to read (and compute syntax of) longer spacing declarations, as well as make it more intuitive to stick to a regular spacing interval.


src/style/functions.scss
```diff
@@ -133,3 +133,8 @@
     @return $colors;
   }
 }
+
+// Scales
+@function spacing-scale($int) {
```
spmcbride1201: Can you rename the variable from `$int` to `$factor` or something? When I see integers, I think "whole numbers without decimal points," but we clearly want to be able to but but any arbitrary value in if it's an integer or not.

shanberg:
@spmcbride1201 Fixed.

shanberg:
@spmcbride1201 My rebase got a little hairy. Would you mind taking another look before I merge?

spmcbride1201:
@shanberg - Of course I'll take a look. No worries.

spmcbride1201:
@shanberg - It seems to work for me, and the source files look sane. Feel free to squash - merge.

PR #367, opened 2017-08-29
Initial Support for Fabric Server

spmcbride1201:
This PR provides initial support for the Fabric Server, a server-side component that provides a consistent API for accessing metrics data.

Given the size of this PR, there are some known regressions that will be handled in subsequent PRs:
- Implementation of Threads support
- Implementation of filtering and auth using attributes returned by `/services` route.
- General cleanup of Jumpstate async logic and addition of tests

Please do not spend much time on the SDS mocking logic. It's just a shim while we wait for the JVM-based backend to be completed.

To start the project:
1. Run `npm run mock-sds`
2. Run `npm start`


.eslintrc
```diff
@@ -10,7 +10,7 @@
     "react/jsx-no-duplicate-props": 1,
     "react/jsx-no-literals": 0,
     "react/jsx-no-undef": 1,
-    "react/jsx-sort-props": 1,
+    "react/jsx-sort-props": 0,
```
spmcbride1201: The straw that broke the camels back was the semantic meaning conveyed by non-alphabetical ordering in React Router v4. It has certain props that can read more like English when ordered like the React Router docs.

public/index.html
```diff
@@ -12,6 +12,7 @@
 
     Valid options include JVM, GOLANG, and ENVOY. Defaults to JVM
   -->
+  <meta property="fabricServer" content="http://localhost:9990">
```
spmcbride1201: I should revert this change and programmatically assume `http://localhost:9990` when the value is unchanged when running in non-production modes

src/components/Container.js
```diff
@@ -38,32 +32,36 @@ class Container extends Component {
...
-              {this.props.runtime === "JVM" &&
-                <Route component={RoutesGrid} path="/route" />}
-              <Route component={GMGrid} path="/dashboard/:dashboardName" />
-              <Route component={NotFound} path="*" />
+              <Route component={SettingsGrid} exact path="/settings" />
+              <Route
+                path="/:serviceName/:instanceID/"
+                component={InstanceRouter}
+              />
+              {/* Just Redirect back to the services page if an instance ID isn't found*/}
```
spmcbride1201: This route is going to eventually provide dashboards with rollups of metrics across the instances of a microservice. It's essentially a placeholder currently.

src/components/InstanceRouter.js
```diff
@@ -0,0 +1,66 @@
...
+import { Actions } from "jumpstate";
+
+class InstanceRouter extends Component {
+  static propTypes = {
+    match: PropTypes.object.isRequired,
+    selectedInstance: PropTypes.string
+  };
+
+  componentDidMount() {
+    Actions.selectInstance(
```
spmcbride1201: As is, the app only polls a single microservice instance at a time. This Jumpstate Effect is responsible for this behavior. It's implemented with these lifecycle hooks to automatically select based on user navigation.

src/components/ServicesGrid.js
```diff
@@ -0,0 +1,63 @@
...
+    Actions.fetchServicesFromServer();
+  }
+
+  setFilterString = filterString => this.setState({ filterString });
+
+  render() {
+    const { services } = this.props;
+    if (services && services.length > 0) {
+      return (
+        <div className="routes-table-container">
```
spmcbride1201: This is reusing SCSS written for the routes table. This should be forked for this component or explicitly generalized

src/components/ServicesTableLineItem.js
```diff
@@ -0,0 +1,144 @@
+import { PropTypes } from "prop-types";
+import React, { Component } from "react";
+import Collapse from "react-collapse";
+import { Link } from "react-router-dom";
+import { trimID } from "../utils";
+
+import IndicatorIcon from "./IndicatorIcon";
+
+export default class ServicesTableLineItem extends Component {
```
spmcbride1201: This component is unfinished. It will be completed in a follow-up PR. Commented out code is preserved to be handled in that follow-on effort.

src/components/ServicesTableToolbar.js
```diff
@@ -0,0 +1,43 @@
...
+        </form>
+      </div>
+      <div className="toolbar-right">
+        <div className="uk-button-group">
+          <button className="btn">
+            <span className="label">Sort</span>
+            <span className="icon" data-uk-icon="icon: triangle-down" />
+          </button>
+          <div data-uk-dropdown="mode: click; pos: bottom-right; boundary: ! .uk-button-group; boundary-align: true;">
+            <ul className="uk-nav uk-dropdown-nav">
```
spmcbride1201: Just mocking the dropdown nav pending further iteration

src/components/SidebarInstance.js
```diff
@@ -0,0 +1,191 @@
...
+  /**
+   * Utility function for generating dynamic dashboards derived from JSON
+   * @param {*} serviceName 
+   * @param {*} instanceID 
+   * @param {*} dashboards 
+   * @param {*} metrics 
+   */
+  generateDynamicDashboards(serviceName, instanceID, dashboards, metrics) {
+    return _.toPairs(dashboards).map(([key, value]) => {
+      const hasValidChart = _.has(value, "summaryCard.chart.type"); // && _.has(dashboard, 'summaryCard.chart.dataAttribute')
```
spmcbride1201: Looks like zombie code.

src/components/SidebarInstance.js
```diff
@@ -0,0 +1,191 @@
...
+    return (
+      <div className="summary-bar">
+        {service &&
+          <SidebarNavWidget
+            currentID={instanceID}
+            parentName={service.name}
+            parentPath={`/${service.name}`}
+            siblings={service.instances}
+          />}
+        {/** Render Services if the path does not contain a selected service or instances*/}
```
spmcbride1201: Both of these comments are old and should be deleted.

src/components/SidebarRouter.js
```diff
@@ -0,0 +1,19 @@
...
+import SidebarServices from "./SidebarServices";
+import SidebarInstance from "./SidebarInstance";
+
+/**
+ * Subrouter for Sidebar
+ * @param {*} props 
+ */
+function SidebarRouter(props) {
+  return (
+    <Switch>
```
spmcbride1201: This might look a bit odd to embed routing logic deep in the component hierarchy. The goal is to independently swap out parts of the UI based on the router and avoid overloading a component with routing conditionals. This seems to be the architecture suggested by React Router v4

src/components/SidebarServices.js
```diff
@@ -0,0 +1,43 @@
...
+class SidebarServices extends Component {
+  static propTypes = {
+    services: PropTypes.array.isRequired
+  };
+
+  render() {
+    const { services } = this.props;
+    return (
+      <div className="summary-bar">
+        {_.map(services, service => {
```
spmcbride1201: I think using lodash's map is superfluous as I'm using _.values in mapStateToProps.

src/jumpstate/fabric.js
```diff
@@ -0,0 +1,16 @@
+import { State } from "jumpstate";
+
+const fabric = State({
+  initial: {
+    groups: {}, //indexed by groupID
```
spmcbride1201: groups is not yet implemented. This might not be required any longer based on proposed changes the Fabric Server API

src/jumpstate/settings.js
```diff
@@ -1,49 +1,70 @@
...
+    selectedService: null,
+    metricsPollingFailures: 0,
+    runtime: !getFabricServer() ? getRuntime() : null,
     threadsFilter: "all"
   },
-  setPollingAsInitialized(state, payload) {
-    return { ...state, pollingHasInitialized: true };
+  // setPollingAsInitialized(state, payload) {
+  //   return { ...state, pollingHasInitialized: true };
+  // },
```
spmcbride1201: This is probably zombie code, but I want to preserve it in an inert state until future work trying to make this dashboard work when a Fabric Server is not present.

src/store.js
```diff
@@ -12,14 +12,96 @@ import logger from "redux-logger";
...
 import metrics from "./jumpstate/metrics";
 import settings from "./jumpstate/settings";
 import threadsTable from "./jumpstate/threadsTable";
 import dashboards from "./jumpstate/dashboards";
-import { getRuntime, getThreadsEndpoint } from "./utils/head";
+import { getRuntime, getThreadsEndpoint, getFabricServer } from "./utils/head";
 
 // Effects / Asynchronous Actions
 
+// Fabric Server Effects
```
spmcbride1201: I'm not super happy with how this code is currently written. I want to reimplement all Hooks using Effects and figure out a sane API for running either with or without a Fabric server

src/workers/ajax.js
```diff
@@ -27,6 +48,13 @@ export function main(message) {
             .then(data => _.omitBy(data, value => Number.isNaN(value)))
         );
       }
+    case "fetchThreadsFromServer":
+      if (!message.serverEndpoint)
+        return Promise.reject("Missing BSDS Endpoint");
```
spmcbride1201: Rename BSDS to "endpoint of Fabric Server"

public/index.html
spmcbride1201: Also, the comment above this line seems to be out of order.

dgoldstein1:
wow the dashboard is looking better every time I see it!

This pr is really big, was there any way you could have broken this down into two? It's hard to catch bugs and focus in on code when there's 1000+ changes. It's also hard to focus in when there are issues to be wrapped up in future PRs (but I know you're under pressure to move fast).

One overall comment is I would add comments to the top of each new file detailing what is does and what the parent component is (it was a little tricky to differentiate between some of the side components off the bat). I might also fill out the type of `param` instead of leaving it at `{*}`

I would also add a few tickets for testing related to this PR if you're not able to get to adding tests for this. 

One bug I found is that if you start up the app and don't touch it for 20+ minutes it gets stale and sluggish (perhaps related to `proxyConsole.js:56 Warning: Hash history cannot PUSH the same path; a new entry will not be added to the history stack`?) but I don't think this is a huge deal. Overall, looks good to me!

PR #378, opened 2017-09-05
Extensible support for language runtimes

spmcbride1201:
This PR attempts to cleanup the architecture for providing support for alternate runtimes. It moves runtime-specific UI components into distinct directories.

To view JVM:
1. Open a terminal session to the project directory and run `npm run mock-jvm`
2. Open `/public/index.html` and ensure runtime is set to `JVM` on line 15
3. Open `package.json` and ensure that proxy is set to `http://localhost:9990`
4. Open a terminal session to the project directory and run `npm start`

image

To view Go: 
1. End all terminal processes used to show JVM
2. Open a terminal session to the project directory and run `npm run mock-go`
3. Open `/public/index.html` and ensure runtime is set to `GOLANG` on line 15
4. Open `package.json` and ensure that proxy is set to `http://localhost:9092`
5. Open a terminal session to the project directory and run `npm start`

You should see something like the following. Notice that this dashboard has Percent CPU utilized, but the JVM dashboard did not. This is because this is a new metric that was not offered by Finagle.
image

## Known Regressions:
* The "drawers" of the Sidebar cards dynamically generated from JSON do not open as expected
* The React Router NavLInks do not update immediately, but hang until the next render loop from React. This suggests that something is wrong in a component lifecycle method.


config/webpack.config.dev.js
```diff
@@ -188,7 +188,8 @@ module.exports = {
           {
             loader: require.resolve("css-loader"),
             options: {
-              importLoaders: 1
+              importLoaders: 1,
```
spmcbride1201: All of the changes in this file are just to enable scss source maps in dev mode.

src/components/AppBrandBar.js
```diff
@@ -6,6 +6,7 @@ const AppVersion = "0.3.3";
 const BrandName = "Fabric";
 const BrandLogo = decipherLogo;
 
+/** Sidebar Branding and Versioning */
```
spmcbride1201: It seems that I've been randomly adding new inline comments. To some degree, this was random, but I plan to go back and have a documentation PR. Since I'm refactoring so much right now and I've got different feature branches that work slightly differently, documentation seems pretty critical

src/components/Container.js
```diff
@@ -38,36 +43,79 @@ class Container extends Component {
...
+    }
+  }
+
   render() {
     return (
       <div id="app-container">
-        <nav className="app-sidebar">
+        {/* Note: the following backgroundImage tag was inlined to get the URL to properly resolve.*/}
+        {/* This is currently resulting in a flicker on initial load, as the inline styles load first.*/}
+        <nav
```
spmcbride1201: The naming of all of these sidebar components is pretty bad. I plan to go back to try to clean it up and refactor into a top level Sidebar component containing.

src/components/DataTotals.js
```diff
@@ -7,6 +7,12 @@ DataTotals.propTypes = {
   sentBytes: PropTypes.number.isRequired
 };
 
+/**
+ * Simple card component that expresses total uploads and downloads
+ * @param {Object} props 
+ * @param {number} props.receivedBytes 
+ * @param {number} props.sentBytes
```
spmcbride1201: I believe that this is the properly way to annotate a complex object passed as an arg. It doesn't seem to feel into VSCode IntelliSense as I expected though, so it could be wrong though.

src/workers/ajax.js
```diff
@@ -27,6 +27,26 @@ export function main(message) {
             .then(data => _.omitBy(data, value => Number.isNaN(value)))
         );
       }
+    // Temporaryily commented out pending further commit
```
spmcbride1201: This is dead code from merging in stuff from `fabric-wide-dash` (a different feature branch that I had to put on hold). Previously, I allowed the scraping of multiple endpoints, but I tightened this down to a single endpoint assuming that we would have access to a backend component. I *may* need to revert to the previous approach of supporting different endpoints, in which case I would use this dead code.

PR #383, opened 2017-09-06
Substantial refactor of Sidebar and Routing

### LINKED ISSUE #382: Regressions with SidebarCards generated from JSON
* The "drawers" of the Sidebar cards dynamically generated from JSON do not open as expected
* The React Router NavLInks do not update immediately, but hang until the next render loop from React. This suggests that something is wrong in a component lifecycle method.

spmcbride1201:
Firstly, this PR solves the sidebar regressions introduced by a previous PR:
* Sidebar SummaryCard open and close their drawers to show detailed metrics when clicked
* Sidebars are highlighted as active based on the current route as expected

Additionally, this PR substantially modularizes the way that the app handles different runtimes. Runtime-specific components are organized into their own folders and have a much more sane and extensible structure. The sidebar has also been refactors into a clear hierarchy of components.

Resolves #382 

## To Test

To view JVM:
1. Open a terminal session to the project directory and run `npm run mock-jvm`
2. Open `/public/index.html` and ensure runtime is set to `JVM` on line 15 (it may already be set)
3. Open `package.json` and ensure that proxy is set to `http://localhost:9990`
4. Open a terminal session to the project directory and run `npm start`

image

If you see fewer cards than expected, please "Reset Dashboards" from the settings page 

To make sure the bugs were fixed:
1. Click on the toggle on one of the sidebar cards. You should see a drawer with metrics open / close.
2. Click on different sidebar cards. The sidebar card should navigate you to a different route and the card should be styled as currently active.

To view Go: 
1. End all terminal processes used to show JVM
2. Open a terminal session to the project directory and run `npm run mock-go`
3. Open `/public/index.html` and ensure runtime is set to `GOLANG` on line 15
4. Open `package.json` and ensure that proxy is set to `http://localhost:9092`
5. Open a terminal session to the project directory and run `npm start`

You should see something like the following. Notice that this dashboard has Percent CPU utilized, but the JVM dashboard did not. This is because this is a new metric that was not offered by Finagle.
image


tilleryd:
@spmcbride1201 I like where this is going.  You may be able to apply another layer of abstraction to the SummaryGrid components by using a HOC they both use.  You could move redundant things like mapping state to props, hostname/port, page title, etc to the HOC.  Then refactor the SummaryGrid components to functions as the HOC would be the only React component required in this pattern.  Certainly not required for this PR but something to consider for future improvements.

shindigira:
**Verified Regressions fixed!**

Using with 'JVM'
screen shot 2017-09-07 at 10 53 53 am
screen shot 2017-09-07 at 10 42 10 am
screen shot 2017-09-07 at 10 42 07 am

Using with `GOLANG`:
screen shot 2017-09-07 at 11 06 56 am
screen shot 2017-09-07 at 11 06 23 am

src/components/Sidebar.js
```diff
@@ -1,164 +1,28 @@
...
+  /* This is currently resulting in a flicker on initial load, as the inline styles load first.*/
+  return (
+    <nav
+      className="app-sidebar"
+      style={{ backgroundImage: `url(${GMFabricBg})` }}
+    >
+      <SidebarHeader />
+      <SidebarNavWidget />
+      <SidebarContent />
+      <SidebarFooter />
```
shindigira: The link is not anchoring to bottom:   
screen shot 2017-09-07 at 4 08 42 pm

spmcbride1201: Good catch. That's definitely a regression.

shindigira:
Is the `Threads` value correct?   
screen shot 2017-09-07 at 4 13 23 pm

src/components/Container.js
```diff
@@ -2,25 +2,19 @@ import { Actions } from "jumpstate";
...
-import RoutesGridJVM from "./jvm/RoutesGrid";
-import SettingsGrid from "./SettingsGrid";
-import Sidebar from "./Sidebar";
-import SidebarFooter from "./SidebarFooter";
-import SummaryGridJVM from "./jvm/SummaryGrid";
-import SummaryGridGolang from "./golang/SummaryGrid";
-import ThreadsGridJVM from "./jvm/ThreadsGrid";
 
-import GMFabricBg from "../images/gm-fabric-bg.jpg";
+import Sidebar from "./Sidebar";
```
shindigira: Yay! More separation of concerns!

spmcbride1201:
The threads sidebar card value definitely looks wrong. I think that's a regression too.

PR #390, opened 2017-09-11
chore: Replace link tag with embedded fonts.

### LINKED ISSUE #209: Embed externally linked fonts into build

spmcbride1201:
The goal of this PR is to embed Google fonts into the project rather than load them via the Internet. This is important because the dashboard will be deployed to environment with slow or non-existent connectivity to 3rd party web services. 

To test this PR, you just need to run `npm start`. I suggest using a Chrome plugin called WhatFont which is available at: https://chrome.google.com/webstore/detail/whatfont/jabopobgcpjmedljpbcaablpmlmfcogm?hl=en

Once the plugin is installed and your restart your browser, you can click on text and see the actual typography being rendered. It should be the following when you click on a title in the sidebar:

image

Resolves #209


tilleryd:
LGTM

PR #391, opened 2017-09-13
Go Dashboards at parity with JVM

### LINKED ISSUE #322: Create new JSON-based dashboards for GM-Fabric-Go
In order to prepare for future capabilities around user-generated custom dashboards, `gm-fabric-dashboard` has migrated a substantial number of dashboards from bespoke custom JSX components to JSON that is passed into a "dynamic dashboard engine" that generates dashboards from a library of GM Fabric UI components.

When looking at `gm-fabric-jvm`, this includes the HTTP, JVM, and Finagle dashboards. Because all dashboards a tied to specific keys, dashboards must be created targeting a particular runtime. 

This task involves modifying the way that GM Fabric uses runtimes to figure out which dashboards to display to be able to render appropriate dashboards for the desired runtime (e.g. making sure JVM dashboards don't render for a Golang microservice and visa-versa), and then designing and building dashboard JSON relevant to the Go runtime.

### LINKED ISSUE #336: Implement custom static dashboards for Go
At minimum, this likely includes the /summary landing page and a /routes landing page. However, this might also include a “Threads” type page (if this is implemented by Doug).

spmcbride1201:
This PR brings the new Go dashboards to functional parity with the JVM.

To test:
1. Clone this project if you don't already have it and checkout the branch `new-go-metrics`
2. Open up `public/index.html` and set the runtime tag on line 15 from
```html
<meta property="runtime" content="__RUNTIME__">
```
to
```html
<meta property="runtime" content="GOLANG">
```
You might be curious what this meta tag does. The dashboard uses this to allow configuration of the app at deploy time, and I have some nifty bash scripts that change these values automatically to make sure the dashboard knows the type of runtime logic to display.

3. Now you need to pull a test go service. This docker container includes three different services that we are going to run the dashboard against. We'll do this by mapping the 1337 port to the three ports of the docker container in sequence. Think of these three ports as different microservices that have been packaged together to make them easier to pull from DockerHub.

Open a new terminal and run
```bash
docker pull drfogout/metricssimple
docker run -it -p 1337:10001 -p 20001:20001 -p 30001:30001 drfogout/metricssimple
```
4. Open a second terminal in the gm-fabric-dashboard directory and run `npm start`
5. Confirm that the dashboard starts cleanly and displays metrics
6. Now we need to stop the container. Normally, on UNIX systems, one is able to hit Ctrl+C to send a SIGINT or signal interrupt to the running process, but in this particular case we don't have these niceties. After hitting Ctrl+C, you'll be back at a clean terminal, but if you run `docker ps` to list the running docker processes (containers), you'll see that drfogout/metricssimple is still running. 

```bash
 ~ docker ps
CONTAINER ID        IMAGE                    COMMAND             CREATED             STATUS              PORTS                                                                         NAMES
b56a40f61a42        drfogout/metricssimple   "/entrypoint.sh"    About an hour ago   Up About an hour    0.0.0.0:10001->10001/tcp, 0.0.0.0:20001->20001/tcp, 0.0.0.0:1337->30001/tcp   pedantic_khorana
```

What the hell, right? Well, in this case, we need to get ultra-violent and run `docker kill`, passing the docker ID from above. 

```bash
 ~ docker kill b56a40f61a42
b56a40f61a42
```

7. Now we are going to map 1337 to the second port
```bash
docker run -it -p 10001:10001 -p 1337:20001 -p 30001:30001 drfogout/metricssimple
```
Once up, we can repeat steps 4-6

8. Now we are going to map 1337 to the third port and repeat steps 4-6
```bash
docker run -it -p 10001:10001 -p 20001:20001 -p 1337:30001 drfogout/metricssimple
```
Once up, we can repeat steps 4-6

All good? Please include comments in the review if you see anything odd in the dashboards. Also contact me if a step doesn't work for you.

Resolves #322 
Resolves #336


stephenfriedman:
Following the tutorial style, I would mention after step one to run `npm install`. 

Following the steps was easy as they were very clear. I was able to get the Dashboards up and running without seeing anything odd.

src/components/golang/FunctionsTableToolbar.js
```diff
@@ -43,12 +43,7 @@ export default function FunctionsTableToolbar({
           <div data-uk-dropdown="mode: click; pos: bottom-right; boundary: ! .uk-button-group; boundary-align: true;">
             <ul className="uk-nav uk-dropdown-nav">
               <li onClick={evt => setKeyToSortBy("func")}>Function Name</li>
-              <li onClick={evt => setKeyToSortBy("inThroughput")}>
-                In Throughput
-              </li>
-              <li onClick={evt => setKeyToSortBy("outThroughput")}>
-                Out Throughput
-              </li>
+              <li onClick={evt => setKeyToSortBy("requests")}>Requests</li>
```
stephenfriedman: Just making sure, this seems to be one of the only place where requestsPerSecond_dygraph and requestPerSecond_sparkline are not used. Is this intentional?

src/components/golang/SummaryGrid.js
```diff
@@ -23,75 +28,78 @@ class SummaryGrid extends Component {
...
                   value={`${getLatestAttribute(
                     metrics,
-                    "route/all/latency_ms.avg",
-                    3,
+                    "all/latency_ms.avg",
+                    5,
                     "ms",
                     "s"
-                  )} s`}
+                  )}s`}
```
stephenfriedman: Is there a reason the 's' is appended here and can't be handled somewhere else? It feels brute forced

PR #398, opened 2017-09-14
Fabric wide dash

spmcbride1201:
This PR is my attempt to rebase all of my previous `fabric-wide-dash` work with my more recent work bringing Go dashboard support up to parity with the JVM.

A significant achievement is that the dashboard can run either with or without a Fabric Server. 

When `index.html` looks like this, the dashboard works as before
```html
<meta property="fabricServer" content="__FABRIC_SERVER__">
```

However, when the endpoint for a Fabric Server is provided, all other `index.html` meta tags are ignored and the dashboard pulls services and metrics information from the Fabric Server
```html
<meta property="fabricServer" content="http://localhost:9990">
```

This should allow the dashboard to be deployed per-instance as before and also in the new manner where a single dashboard monitors an entire fabric. 

To test running against a Fabric server:
`npm run mock-sds` and then configure the fabric server meta tag

To test running directly against a Go microservice without a Fabric server:

```bash
docker pull drfogout/metricssimple
docker run -it -p 1337:10001 -p 20001:20001 -p 30001:30001 drfogout/metricssimple
```
 and then configure the runtime meta tag in `index.html`. Make sure the fabric server meta tag is not set.

Known Issues:
- Fabric Server mode does not yet load threads metrics from the Fabric Server when the threads component mounts
- The `mock-sds` mock service does not yet mock go metrics
- Groups and other aspects of the Fabric Server have not been implemented
- Per agreement with the backend engineers, instances is not included in the response to the GET /services route. The `mock-sds` endpoint will have to be modified in anticipation of these changes.


src/components/SidebarNavWidget.js
```diff
@@ -1,33 +1,96 @@
...
+          <span
+            className="icon"
+            data-uk-icon={`icon: chevron-left; ratio: 1`}
+          />
+          <span className="label">{parent.name}</span>
+        </a>
+      )}
+      {parent &&
+      parent.name &&
+      parent.path &&
```
tilleryd: https://lodash.com/docs/#get could dry this up with `_.get(parent, ['name', 'path']);`

src/components/SidebarServices.js
```diff
@@ -0,0 +1,44 @@
...
+
+import SidebarCard from "./SidebarCard";
+
+class SidebarServices extends Component {
+  static propTypes = {
+    services: PropTypes.array.isRequired
+  };
+
+  render() {
+    const { services } = this.props;
```
tilleryd: 1 time use var

src/store.js
```diff
@@ -13,25 +13,98 @@ import logger from "redux-logger";
...
+  const fabricServer = getState().settings.fabricServer;
+  if (fabricServer) {
+    Actions.fetchMetricsWithServer(arg);
+  } else {
+    Actions.fetchMetricsWithoutServer(arg);
+  }
+});
+
+/**
+ * Async Jumpstate Effect that instructs the AJAX worker to fetchMetrics and calls success or failure actions
```
tilleryd: fetchMetricsWithoutServer

src/store.js
```diff
@@ -72,23 +195,95 @@ Effect("fetchThreadsFailure", err => {
...
+Effect("startPollingWithoutServer", function({
+  endpoint = getState().settings.metricsEndpoint,
+  interval = getState().settings.interval
+}) {
+  console.log("startPollingWithoutServer", endpoint, interval);
+  // We need to make sure we clear any existing
+  if (window.refreshMetricsIntervalID) {
+    clearInterval(window.refreshMetricsIntervalID);
+    window.refreshMetricsIntervalID = null;
+  }
```
tilleryd: could make this a utility function as it is copied in 3 places

tilleryd:
I now see what you mean about modularizing the service instance components.  We certainly have an opportunity to reduce some of the redundant code in `instances/golang` and `instances/jvm`.  In `RoutesGrid` components, the only difference is the import of `getRoutesTable`.  An HOC that conditionally `requires` the module could be a quick solution.  `SummaryGrid` components are quite different from each other but still have some similarities that can factored out.

tilleryd:
I quickly tested the dashboard this morning using the mock JVM service and all seems to be working.  Yesterday I tested against the mock SDS and Doug's Go microservice.  You did a great job of maintaining stability despite the massive size of this PR.  

Excellent work!  LGTM 👍

src/store.js
spmcbride1201: What's the feedback here?

tilleryd: Just that the comment says `fetchMetrics` instead of `fetchMetricsWithoutServer`.

PR #411, opened 2017-09-20
Implement Service Polling

### LINKED ISSUE #291: Implement polling of Service Discovery Service to populate Fabric services data in Redux
The dashboard needs access to an API or templated JSON endpoint containing the state of the Fabric. This data should be fed into Redux and used to populate nav elements and configure how the dashboard can get access to instance-specific metrics (either via polling of scrape-points or querying of a TSDB)

spmcbride1201:
This PR implements polling of services from a mock of the discovery-service.

Additionally, it accomplished the following:
- It updates the mock discovery service to reflect recent API changes.
- It modifies Go runtime components to run properly when running under `discovery-service` mode.
- It cleans up the directory structure of the `mock-json` components
- It added a handful of unit tests

Resolves #291


src/components/App.js
```diff
@@ -106,13 +116,20 @@ function mapStateToProps({
   dashboards,
   fabric: { services },
   routing: { location: { pathname } },
-  settings: { interval, metricsEndpoint, runtime, selectedService }
+  settings: {
+    instancePollingInterval,
+    metricsEndpoint,
+    runtime,
+    selectedService,
+    selectedServiceVersion
```
tilleryd: This would be a good opportunity to use `reselect` so you can just use 1 prop `selectedServiceVersion` in your render function.

```
const getSelectedServiceVersion = createSelector(
  state => state.settings.selectedService,
  state => state.settings.selectedServiceVersion,
  (selectedService, selectedServiceVersion) =>
    `${selectedService}${selectedServiceVersion}`
);

function mapStateToProps(state) {
  return {
    selectedServiceVersion: getSelectedServiceVersion(state)
  };
}
```

You'll have to massage this example so you can plug it into your object destructuring of state props.

src/components/instance/jvm/SummaryGrid.js
```diff
@@ -41,7 +43,8 @@ class SummaryGrid extends Component {
     return (
       <div>
         <PageTitle
-          title={`${selectedService || getServiceName()}: ${trimID(
+          title={`${selectedService ||
+            getServiceName()} ${selectedServiceVersion} : ${trimID(
```
tilleryd: Since this code is used in both JVM and Go `SummaryGrid` components then perhaps store `selectedServiceTitle` in Redux.

src/utils/index.test.js
```diff
@@ -0,0 +1,18 @@
...
+  });
+  test("uses INSTNACE_ID_LENGTH when desired length is not provided", () => {
+    expect(trimID("Decipher Tech Studios")).toEqual(
+      trimID("Decipher Tech Studios", INSTANCE_ID_LENGTH)
+    );
+  });
+  test("returns the entire ID when desired length is longer than the length of the actual ID", () => {
+    expect(trimID("Decipher", 20)).toEqual("Decipher");
+  });
+});
```
tilleryd: Witness the boost in code coverage!!

src/components/App.js
tilleryd: Actually, something came to mind to me this morning.  You don't need `reselect`, just set a computed property `selectedServiceVersion` in the store.  `reselect` would be way overkill in this use case.

src/components/instance/jvm/SummaryGrid.js
spmcbride1201: This kinda brings is a bunch of rework around how my util functions for head meta tags are called throughout my app and the conflict between using the meta tags and the Redux store as competing sources of truth. Since reselect functions need to be pure, I'm going to have to table this as a new issue.

tilleryd: Understood

PR #419, opened 2017-09-27
405 implement list component

karinakeen:
Work in progress: add stateless section list view component.


karinakeen:
implement dashboard gateway page lists and cards

PR #422, opened 2017-09-29
421 integrate toolbar components

karinakeen:
This PR integrates toolbar and card / list view items.


json-mock/discovery-service/data.js
```diff
@@ -32,6 +32,91 @@ const services = [
       "8bedb4551e801f38bf149000002a1127",
       "d000009c26c6c84daaf1ceb40559d659"
     ]
+  },
+  {
+    name: "Whacky Hidden Missile Silo",
```
spmcbride1201: This is just the best!

PR #428, opened 2017-10-02
399 services gateway sidebar

### LINKED ISSUE #399: Services Gateway sidebar
### Accordions
- Buckets of services states
  - Vertically ordered top/down by: Down, Warning, Stable
  - Header displays the total number of services per status

## Functional requirements
- Tabbing through the accordions should work as expected to support a11y.
- When an accordion is expanded then tabbing should step through the list elements before tabbing to adjacent accordions.

karinakeen:
This PR implements sidebar view displaying all services in fabric.  

Resolves #399


src/utils/selectors.js
```diff
@@ -167,3 +169,37 @@ function _buildRoutesTree(routeMetrics) {
...
+      state: computeState(
+        service.instances.length,
+        service.minimum,
+        service.maximum
+      )
+    };
+  })
+);
+
+function computeState(count, min, max) {
```
spmcbride1201: Can you please add JSDocs documentation here? I think it should be pretty quick and easy with the plugin I had you install. If you right click right above 194, you can select "document this"
image

That should create documentation, but the type annotations will be "any". I think these should all be "number" instead.

Here's the resource I refer to when documenting this:
http://usejsdoc.org/tags-param.html

src/components/fabric/FabricSidebarContentSectionItem.js
```diff
@@ -0,0 +1,41 @@
...
+
+// styled components
+const Container = styled.div`
+  color: white;
+  display: flex;
+  align-items: center;
+  padding: 5px 0 5px 0;
+`;
+
+const Box1 = styled.span`
```
spmcbride1201: Can we rename this to something more descriptive like "TitleSpan" and "VersionSpan"?

src/components/fabric/FabricSidebarContentSection.js
```diff
@@ -0,0 +1,120 @@
...
+              />
+            </Collapse>
+          );
+        })}
+      </div>
+    );
+  }
+}
+
+function getColor(state) {
```
spmcbride1201: Please add JSDocs commenting. I discuss how to do this below (I guess I'm reviewing your code from bottom to top for some weird reason).

src/components/fabric/FabricSidebarContentSection.js
```diff
@@ -0,0 +1,120 @@
...
+  let color;
+
+  switch (state.toLowerCase()) {
+    case "warning":
+      color = "#ffcc00";
```
spmcbride1201: It looks like color is a variable that we really don't need. This looks like each case should just return a string directly:
```js
case "warning":
  return "#ffcc00";
```

src/components/fabric/FabricSidebarContentSection.js
```diff
@@ -0,0 +1,120 @@
...
+const Header = styled.div`
+  color: white;
+  display: flex;
+  background-color: #333333;
+  align-items: center;
+  justify-content: space-between;
+  border-bottom-width: 2px;
+  border-radius: 5px;
+  border: 2px solid #333333;
+  border-bottom-color: ${props => props.borderBottomColor};
```
spmcbride1201: Great use of props!!!

src/components/fabric/FabricSidebarContentSection.js
```diff
@@ -0,0 +1,120 @@
...
+`;
+
+const HeaderPart1 = styled.span`min-width: 75%;`;
```
spmcbride1201: `HeaderLeft` and `HeaderRight`, maybe?

src/components/fabric/FabricSidebarContentSection.js
```diff
@@ -0,0 +1,120 @@
...
+          <HeaderPart2>
+            {services.length > 1 && <ItemCount>{services.length}</ItemCount>}
+            <Down>
+              <TriangleDown fill="white" stroke="white" />
+            </Down>
+          </HeaderPart2>
+        </Header>
+
+        {_.map(services, ({ name, state, version, docsLink }) => {
+          docsLink = docsLink || "example.com";
```
spmcbride1201: So if you want to set a default, I think the best way with ES6 is to use [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters), which looks like this:
```js
{_.map(services, ({ name, state, version, docsLink = "example.com" }) => {
```

However, I don't think we need to do this here, as we probably don't want to render a link at all if docsLink isn't provided. I think a reasonable way of doing this would be to pass a falsy value through to `FabricSidebarContentSectionItem`.

src/components/fabric/FabricSidebarContent.js
```diff
@@ -0,0 +1,108 @@
...
+import styled from "styled-components";
+
+// styled components
+
+const Services = styled.div`
+  padding: 10px 0 10px;
+  background-color: #666666;
+  border-radius: 5px;
+`;
+const Header = styled.div`
```
spmcbride1201: It looks like there might be some identical styled-components in FabricSidebarContent and FabricSidebarContent section. If that's the case, can we just have one copy and import from one into the other?

karinakeen:
Added documentation, renamed styled components, cleaned getColor function

PR #440, opened 2017-10-03
424 status SVG

spmcbride1201:
This PR is a first pass at #424.


src/components/fabric/FabricSidebarContentSection.js
```diff
@@ -41,20 +53,20 @@ export default class FabricSidebarContentSection extends Component {
...
-              <IndicatorIcon color={"lightblue"} diameter={12} />
-            </HeaderIcon>
-            {header}
+            <IconWrapper>
+              <StatusIcon status={status} />
+            </IconWrapper>
+            {status}
           </HeaderLeft>
           <HeaderRight>
             {services.length > 1 && <ItemCount>{services.length}</ItemCount>}
```
karinakeen: {services.length > 1 && <ItemCount>{services.length}</ItemCount>}
change to "services.length > 0..." ItemCount is not called when there is only one service.

PR #444, opened 2017-10-04
425 make service gateway search router driven

karinakeen:
This PR:
Gateway search is now router driven with a single query parameter.  Add multiple query parameters...


public/index.html
```diff
@@ -14,8 +14,8 @@
     value is anything other than __FABRIC_SERVER__, the app runs in Fabric Server
     mode and the attributes for "running without a Fabric Server" are ignored
   -->
-  <!-- <meta property="fabricServer" content="http://localhost:1337"> -->
-  <meta property="fabricServer" content="__FABRIC_SERVER__">
+  <meta property="fabricServer" content="http://localhost:1337">
```
spmcbride1201: Can you please revert this change? Line 17 should be commented out and like 18 should be uncommented.

src/components/fabric/FabricGrid.js
```diff
@@ -3,55 +3,104 @@ import { PropTypes } from "prop-types";
...
+
+    // Debounce
+    this.updateUrl = _.debounce(this.updateUrl, 500);
   }
+
+  componentWillMount() {
+    // Parse the current query parameter
+    const queryParam = qs.parse(this.props.location.search) || "";
+
+    // maybe rewrite
```
spmcbride1201: Is this comment still relevant or can it be deleted?

src/components/fabric/FabricGrid.js
```diff
@@ -3,55 +3,104 @@ import { PropTypes } from "prop-types";
...
+   */
+  onChange = query => {
+    this.setState({ query });
+
+    // updateUrl() is debounced in constructor.
+    this.updateUrl();
+  };
+
+  /**
+   * updateUrl function takes the saved input and sets the query.
```
spmcbride1201: Can you update this comment to reflect the fact that we now only have query?

PR #447, opened 2017-10-04
fix: resolves issue 445

karinakeen:
This PR:

Add "babel-polyfill" to dependencies
Requires "babel-polyfill" at the top of the entry point in webpack.config.dev.js and webpack.config.prod.js
Removes "object-assign" from dependencies
Removes "object-assign" from polyfills.js

Resolves issue #445


public/index.html
```diff
@@ -14,8 +14,8 @@
     value is anything other than __FABRIC_SERVER__, the app runs in Fabric Server
     mode and the attributes for "running without a Fabric Server" are ignored
   -->
-  <!-- <meta property="fabricServer" content="http://localhost:1337"> -->
-  <meta property="fabricServer" content="__FABRIC_SERVER__">
+  <meta property="fabricServer" content="http://localhost:1337">
```
spmcbride1201: Can you reverse these?

PR #451, opened 2017-10-04
Style Elements

spmcbride1201:


spmcbride1201:
Created PR on behalf of @shanberg. I'm unable to formally review, but I went through the code and it looked good after resolving some linting issues.

PR #510, opened 2017-10-09
481: dashboard header alignment issue

### LINKED ISSUE #481: Dashboard IE11 view alignment issue:  The dashboard has headers that overlap with metrics
The dashboard has headers that overlap with metrics

IE11

screen shot 2017-10-05 at 3 00 10 pm

karinakeen:
Resolves #481


src/components/library/ReadoutItem.js
```diff
@@ -1,6 +1,11 @@
 import React from "react";
 import { PropTypes } from "prop-types";
 
+// add minimum height to flex items. `flex-direction:column` based flex container will use `min-height` to calculate `flex-grow` lengths in IE11.
+const minHeight = {
+  minHeight: "1em"
```
spmcbride1201: This should be fine to move into SCSS. I think lines 122-132 in `data-displays.scss` is the correct place, but you should probably double check. Thanks!

PR #513, opened 2017-10-09
feat: Implement fetching threads from discovery service

### LINKED ISSUE #370: Implement Threads API support

spmcbride1201:
This PR implements support for fetching metrics from the server-side discovery service.

To see the threads data, open a JVM-based microservice (e.g. Batcomputer) instance and select Threads from the sidebar.

image

Resolves #370


spmcbride1201:
@karinakeen - I clarified the logic with inline comments and better variable names. Let me know if this clarifies what's going on. It's fine if it's still unclear, as we want to make sure this codebase is approachable for people looking at it for the first time.

PR #514, opened 2017-10-09
chore: Adjust layout for medium width

### LINKED ISSUE #474: Dashboard view alignment issue:   Tables don't align vertically on Go (Both Chrome and IE11)
Dashboard view - Go (Both Chrome and IE11)

Tables don't align vertically when window is resized.

screen shot 2017-10-05 at 2 41 04 pm

spmcbride1201:
This PR adjusts JSON settings used to generate a few of the dashboards so that the charts scale properly when displayed at the medium breakpoint. In addition to the issues with the Go dashboard, this also fixes an identical issue for the JVM dashboard for JVM microservices.

Resolves #474


spmcbride1201:
I've seen what you're describing before (where it seems to load at half-width squished to the left). Usually it seems to be for a single render and it resolves on the second render. For whatever reason, I can recreate it right now, but I think the issues is probably outside fo the context of this JSON, so I'll open a separate issue. Unless the dashboards get stuck in that squished position for a prolonged period of time, I think we should merge this.

PR #515, opened 2017-10-09
502 root routes

### LINKED ISSUE #502: Dashboard not showing metrics for "/" route properly.
The requests per second is always 0 and all values are 0'd

spmcbride1201:
Resolves #502


src/components/instance/go/RoutesTableLineItem.js
```diff
@@ -36,10 +36,12 @@ export default class RoutesTableLineItem extends Component {
...
-    const errorPercent =
-      (1 -
-        (this.props.requests - this.props.errorsCount) / this.props.requests) *
-      100;
+    const errorPercent = this.props.requests
+      ? (1 -
+          (this.props.requests - this.props.errorsCount) /
+            this.props.requests) *
+        100
+      : 0;
```
karinakeen: Possible edge cases? when errorsCount > requests, errorPercent > 100%...

spmcbride1201: This shouldn't really be possible if the backend API is working correctly. The requests to the microservice will produce responses, which will have success and errors. In the worst case, every request would result in an error, which would make requests equal errorsCount

karinakeen: Ok.

PR #551, opened 2017-10-11
524 fix inspector search

### LINKED ISSUE #524: Fix use of stateful URLs in Inspector Search
Based on the work implementing stateful URLs for the searching of services, it's clear that inspector search is not implementing the proper pattern for client-side searching and stateful URLs with React Router v4.

The main problem is that the value in the search bar is not updated when the user clicks the back button or does a hard refresh. Data flows unidirectionally from the search field, not conditionally bidirectional as with the fabric search component.

spmcbride1201:
This PR brings the approach I came up in searching for services to searching metrics.

To test this, go to Explorer for a microservice

The following should work:
- Enter a search term. You should be able to type naturally, and at a natural pause in typing, the search word or phrase should be appended to the URL as a search query.
- Enter a second search term. You should see things update and it should replace the existing search query
- Select a metric to open a chart. you should see a second key-value pair get added to the query string in the URL
- Do a hard refresh of the browser, the app should load exactly where you left off
- Click the back and forward buttons. The state should update to reflect what you've done

Resolves #524


src/components/Explorer.js
```diff
@@ -82,91 +82,151 @@ const MetricsGraphDisplay = styled.div`
...
+    // is which the app router action was POP, which means the user hit the back button or
+    // otherwise navigated using the client-side router history
+    if (
+      nextProps.location.search !== `?${this.state.lastPushedQueryString}` ||
+      nextProps.history.action === "POP"
     ) {
-      return true;
-    } else {
-      return false;
+      this.popAndDecodeHistory(nextProps.location.search);
```
kaitmore: Nice solution!

Issue #368, opened 2017-08-29
Refactor SCSS used by Services Table


spmcbride1201:
Currently the `.routes-table-` prefixed styles are used in:
- Fabric Table
- Go Functions Table
- Go Routes Table
- JVM Routes Table

There is also a JVM Thread Table that might use distinct styles, but it very similar.

Once common styled-components are implemented across these tables, create an issue to identify the possibility of creating a common GMTable molecule that is sufficiently extensible to cover each of these use cases.

karinakeen:
blocked by #533

PR #553, opened 2017-10-12
538 fabric dashboard routes trailing error percent

### LINKED ISSUE #538: fabric dashboard routes showing trailing error percent
screen shot 2017-10-11 at 11 58 05 am

karinakeen:
Resolve #538


src/components/instance/go/FunctionsTableLineItem.js
```diff
@@ -34,15 +34,15 @@ export default class FunctionsTableLineItem extends Component {
...
-            this.props.requests) *
-        100;
-    }
+    let errorPercent = this.props.requests
+      ? Math.round(
+          (1 -
+            (this.props.requests - this.props.errorsCount) /
+              this.props.requests) *
+            100 *
+            1000
```
spmcbride1201: I think a better approach for this is `toLocaleString`. It has a bunch of options for setting the maximum number of decimal points, and it provides some basic internationalization support (e.g. in Germany, commas are decimals and decimals are commas).

If we pass undefined for the first argument, the browser will inherit the language locale settings from the operating system. I think the example below for force three decimals at all times.

```js
num.toLocaleString(undefined, {minimumFractionDigits: 3, maximumFractionDigits: 3})
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

src/components/instance/go/FunctionsTableLineItem.js
```diff
@@ -85,7 +85,7 @@ export default class FunctionsTableLineItem extends Component {
         <div
           className={
             "routes-table-error-percent routes-table-monospace err-pc-" +
-            errorPercent
+            parseInt(errorPercent / 10, 10)
```
spmcbride1201: Isn't this already a number? I'm not really seeing a case where this would ever be a string?

src/components/instance/go/RoutesTableLineItem.js
```diff
@@ -36,12 +36,25 @@ export default class RoutesTableLineItem extends Component {
...
-          (this.props.requests - this.props.errorsCount) /
-            this.props.requests) *
-        100
+    // round off any float points after three decimal places.  integers will not have trailing zeros
+    let errorPercent = this.props.requests
+      ? Math.round(
+          (1 -
+            (this.props.requests - this.props.errorsCount) /
+              this.props.requests) *
+            100 *
```
spmcbride1201: Ditto comment for toLocaleString as above.

src/components/instance/go/RoutesTableLineItem.js
```diff
@@ -36,12 +36,25 @@ export default class RoutesTableLineItem extends Component {
...
+            1000
+        ) / 1000
       : 0;
+
+    // roundsTable.scss calculates the error percentage color based on the percentage value.  the mix function in scss takes value of percentage so it will be on the green scale up to 30%, brown between 30%~60%, then get closer to red above 70%.
```
spmcbride1201: Let's try to keep lines under like 80 - 100 characters long so that people don't have to scroll their editors left and right

spmcbride1201: Good comment though. This is really useful to know!

spmcbride1201: Oh, I see that you're using number 0-10 instead of 0-100 now? Can you update the comments to reflect this?

src/components/instance/go/RoutesTableLineItem.js
```diff
@@ -87,7 +100,7 @@ export default class RoutesTableLineItem extends Component {
         <div
           className={
             "routes-table-error-percent routes-table-monospace err-pc-" +
-            errorPercent
+            parseInt(errorPercent / 10, 10)
```
spmcbride1201: Sort of off topic, but I think that we probably want to be a bit more sensitive about the error rate. If something has like a 10% failure rate, I would expect it to be flagged red. I'm not sure about how we'd want to set the thresholds, but maybe < 0.1% error rate is green, > 0.1% and < 1% is yellow, and >1% is red. It's like a logarithmic scale when we extract out the range further.

src/components/instance/go/RoutesTableLineItem.js
```diff
@@ -100,7 +106,7 @@ export default class RoutesTableLineItem extends Component {
         <div
           className={
             "routes-table-error-percent routes-table-monospace err-pc-" +
-            parseInt(errorPercent / 10, 10)
+            Math.round(errorPercent / 10)
```
karinakeen: Maybe open a different issue for the errorPercent coloring?

src/components/instance/go/FunctionsTableLineItem.js
karinakeen: Tweaked it to take the tenth digit for now, but I should create another issue under #368.

src/components/instance/go/RoutesTableLineItem.js
spmcbride1201: Good idea. The new issue is https://github.com/DecipherNow/gm-fabric-dashboard/issues/565

PR #555, opened 2017-10-12
490 a11y selects

### LINKED ISSUE #490: Unable to tab into the Group By / Sort By dropdowns
A user is able to tab across the Group and Sort buttons and activate the dropdown, but there is no way to select an item from the dropdown using the keyboard. I would expect to be able to arrow up or down once the dropdown opens and have focus shifted to the currently selected item in the dropdown.

Note that the dropdown is currently provided by UIKit, which we are trying to refactor out of the application. If it's possible to fix this issue while refactoring the dropdown functionality, even better.

image

spmcbride1201:
Resolves #490


src/style/globals.js
```diff
@@ -0,0 +1,35 @@
...
+        line-height: 28px;
+      }
+    }
+
+    .Select-input {
+      height: 28px;
+    }
+  }
+  
+`;
```
kaitmore: Cool, looks good. If you wanted to get away from classnames, I think you could also pass `Select` into the styled() factory call to override these styles.

PR #562, opened 2017-10-12
557 layoutsection refactor

### LINKED ISSUE #557: Refactor `layout-section.scss` into styled-components

kaitmore:
Resolves #557


src/components/Inspector.js
```diff
@@ -39,6 +39,12 @@ export default class Inspector extends Component {
                   : ""}`}
                 key={key}
                 onClick={evt => onClick(evt.target.innerText)}
+                onKeyDown={evt => {
```
spmcbride1201: 💪💪💪💪💪 Keyboard Strong 💪💪💪💪💪

src/components/library/LayoutSection.js
```diff
@@ -8,28 +11,51 @@ LayoutSection.propTypes = {
...
  * Section of a static dashboard, complete with header and icon
  * @param {Object} props - refer to propTypes 
  */
-function LayoutSection({ children, className, title, icon }) {
+
+function LayoutSection({ children, className = "", title, icon }) {
   return (
-    <section className={"layout-section " + className}>
-      <header>
+    <LayoutSectionWrap className={"layout-section " + className}>
```
spmcbride1201: I noticed that the "layout-section" classname is still used here. Additionally, it seem like all of the uses of the className prop should just extend your new LayoutSectionWrap styled-component. The only place I see that prop getting used currently is in `src/components/SettingsGrid.js`, where one use is active and one is commented out. For the commented out instance, I suggest adding a comment to the effect that the className props has been deprecated and needs to be refactored.

kaitmore: #570

PR #563, opened 2017-10-15
501 fix: show zero instead of negative value when number of requests decr…

### LINKED ISSUE #501: Dashboard graphs showing negative values after services restart
If a service has been restarted after being hit _n_ number of times, the graph of requests per second will show negative values.

karinakeen:
…ease on dashboard graphs

Resolve #501


spmcbride1201:
@karinakeen - After talking things over with @joshua-rutherford, I don't want to merge this into the dashboard. The error condition that this targets theoretically shouldn't be possible, so if it does somehow come up, I want it to be super obvious so we can better do root cause analysis and open issues on the backend. In any case, thanks for your work on this!

PR #599, opened 2017-10-18
594: Card Icon Link fix

### LINKED ISSUE #594: Icon currently blocks the click/touch area for links

kaitmore:
I ended up not refactoring this. To make the icon into a background image like I wanted to originally, I would have had to create duplicate svg files which seemed silly. I tried creating a function that would render the svg so that we could just pass it dimensions but React wasn't having it. 

So anyway, z-index.

Resolves #594


spmcbride1201:
@karinakeen - Please review this one for me. We want to make sure that the icon no longer blocks the link in either IE11 or Chrome.

PR #615, opened 2017-10-18
566 enhance: use toLocaleString in getErrorRate selector to  render to th…

karinakeen:
…ree decimal places

I did not rewrite the logic as the alternatives weren't much better in terms of if/else optimization.


src/utils/jvm/selectors.js
```diff
@@ -10,21 +10,36 @@ import { getMetrics, getRoutesTree, getRoutesMetrics } from "../selectors";
...
+    ).toLocaleString(undefined, {
+      maximumFractionDigits: 3,
+      minimumFractionDigits: 3
+    });
   } else {
-    return Number(0).toFixed(4);
+    return Number(0).toLocaleString(undefined, {
+      maximumFractionDigits: 3,
+      minimumFractionDigits: 3
+    });
```
kaitmore: Looks great! If anything maybe the 2nd argument for toLocaleString() or even the Number(0) return statements could be factored out into variables but not a big deal. :sunglasses:

PR #624, opened 2017-10-18
fix: icons now have the correct opacity in IE

kaitmore:


karinakeen:
Confirmed that the background icons have the correct opacity in IE.  

I noticed two minor things in IE.    
1/ Noticed that the service name on the card is moved to the center instead of top in IE.
2/ Stable background icon is slightly cut off on the right and left.

screen shot 2017-10-18 at 4 42 50 pm

screen shot 2017-10-18 at 4 43 58 pm

Issue #607, opened 2017-10-18
Main content area shows focus border on top and bottom when a user clicks on any child elements


spmcbride1201:
Notice the grey focus border just below the header and just above the footer

![image](https://user-images.githubusercontent.com/8700322/31736563-23cb185c-b413-11e7-8d0c-4123a7cbb15c.png)

kaitmore:
Resolved by #652

PR #652, opened 2017-10-19
607 removed content border

kaitmore:
Now if you tab to the skip navigation button and hit enter, it will bring focus to the main content div without the outline. If you continue to tab it should go to the first child element in the main content area.


kaitmore:
Resolves #607

PR #666, opened 2017-10-23
656 make uptime counter rerender every second

### LINKED ISSUE #656: Make uptime counter update every second regardless of polling rate.
This is an extension of #572. The idea is to have the uptime counter rerender every second to behave like a digital clock, rather than on every poll. Check out this video for similar example code. (https://www.youtube.com/watch?v=EGzh7K1qLwE&utm_source=reactnl&utm_medium=email), but be sure to make sure the setInterval only occurs once and only once per second. This probably means that componentShouldUpdate should only run when the internal state is updated, and the setInterval should only run once per second.

karinakeen:
This PR:

Replace PrettyMS with custom helper function ConvertMS
Re-render fabric dashboard uptime every second
Sidebar uptime is not updated as this feature may become irrelevant. 
**We can now remove PrettyMS completely from webpack and package.json **

Resolve #656


src/utils/index.js
```diff
@@ -34,3 +34,29 @@ export function clearFabricIntervalIfNeeded() {
     window.refreshFabricIntervalID = null;
   }
 }
+/**
```
spmcbride1201: This function is nice. Thanks for the documentation!

src/components/instance/go/SummaryGrid.js
```diff
@@ -32,14 +29,51 @@ class SummaryGrid extends Component {
     selectedService: PropTypes.string,
     selectedServiceVersion: PropTypes.string
   };
+
```
spmcbride1201: This logic looks just like what I had envisioned. I haven't run it yet, but it looks fantastic!

src/utils/index.js
```diff
@@ -34,3 +34,29 @@ export function clearFabricIntervalIfNeeded() {
...
+  let s = Math.floor(ms / 1000);
+  let m = Math.floor(s / 60);
+  s = s % 60;
+  let h = Math.floor(m / 60);
+  m = m % 60;
+  let d = Math.floor(h / 24);
+  h = h % 24;
+
+  [d, h, m, s] = [d, h, m, s].map(el => {
+    if (el === 0) return "00";
```
spmcbride1201: After looking at this in the app, I think that we should only add the units that have values. If we have a value of 1000ms, that should produce `1 s` or perhaps `01 s`, but not `Avg. Response Time 00d 00h 00m 01s`

src/components/Main/scenes/Instance/scenes/jvm/scenes/Summary.js
```diff
@@ -58,19 +95,15 @@ class SummaryGrid extends Component {
...
               </Readout>
 
               <Readout type={"readout-primary"}>
                 <ReadoutItem
                   icon={"bolt"}
                   title={"Avg. Response Time"}
-                  value={prettyMS(
-                    _.round(getLatestAttribute(metrics, "time/2XX.avg"), -3)
-                  )}
+                  value={convertMS(getLatestAttribute(metrics, "time/2XX.avg"))}
```
spmcbride1201: I think that this should remain in milliseconds because this is the average latency of a successful response. That's probably going to be less that 1s normally, so milliseconds is appropriate here.

spmcbride1201: Please also change the Go summary dashboard to express "Avg. Response Time" in milliseconds. It's currently converting to seconds there.

karinakeen: this is done

src/utils/index.js
karinakeen: related to issue #687.  I will add a fix for this in the issue.

PR #671, opened 2017-10-23
Add App Header Components to Storybook

kaitmore:
- Built out the new `AppHeader` in storybook
- Restructured all stories into their own files to avoid an unwieldy `index.js`
- Changed any instances of `COLOR_SIDEBAR` variables to `COLOR_ALT` to broaden its usability
- Created a `globalPatterns.js` that will hold commonly used styles across the app such as active, focus, and hover states, or patterns like hiding an element.


shanberg:
Resolves #664

PR #693, opened 2017-10-25
657 error boundary components ui areas

kaitmore:
We created an `ErrorBoundary` class that is able to render a custom fallback UI via props. It was applied to major UI areas.


src/components/library/ErrorBoundary.js
```diff
@@ -0,0 +1,41 @@
...
+        <div className="content">
+          <span data-uk-icon="icon: warning; ratio: 1.8" />
+          <span>{`Error: ${this.state.info}`}</span>
+        </div>
+      </div>
+    );
+  };
+  // If there is an error, render the fallback UI, else render children
+  render() {
+    return this.state.error ? this.fallBackUI() : this.props.children;
```
spmcbride1201: Nice use of render props!

PR #708, opened 2017-10-26
706 Authorization Bug Fix

### LINKED ISSUE #706: Authorization Redirect always fires on hard refresh
When a user goes to a route and issues a hard refresh, the user is immediately redirected to the root route and presented with a "You are not authorized." message. This includes microservices that the user actually is authorized to see.

image

kaitmore:
Upon navigating to an authorized service and issuing a hard refresh, you should no longer be redirected to the fabric grid. 

If you navigate to an unauthorized route such as `Nachos/1.0`, you should be redirected. If you navigate to an unauthorized route multiple times during a session, it should not show multiple notifications.

Resolves #706


7ruth:
@kaitmore not sure what I am doing wrong, but navigating to http://localhost:3000/#/Nachos/1.0 shows me an errorBoundary message and does not redirect me anywhere. 

Hard refresh does not redirect as expected.

PR #709, opened 2017-10-26
527 metrics trimming

### LINKED ISSUE #527: Implement "Garbage Collection" for the in-memory metrics store in Redux
This should involve finding a way to monitor the memory footprint of this state, determining when a threshold is reached, and then implementing some sort of method of keeping the metrics under the threshold.

spmcbride1201:
The purpose of this PR is to trim the metrics store in Redux to prevent it from growing too large and crashing the app. The code currently includes a bunch of console logs which might help you understand what's going on. I've also set the max size to 500k to speed up how long if will take for you to see the trimming functionality start. I suggest selecting a JVM microservice and leaving the Summary page open for a prolonged period of time to see the UI behavior, which should appear as the oldest leftmost data flowing off the line chart 

After approval, I will clear out the console logs and increase the max metrics size to 100MB.

Resolve #527


src/jumpstate/metrics.js
```diff
@@ -1,15 +1,41 @@
...
+
+const METRICS_CACHE_MAX_BYTES = 500000; // ~100MB is 100000000
 
 // State Objects
 const metrics = State({
   initial: {},
   setMetrics(state, payload) {
-    const result = { ...state };
-    const timestamp = Date.now() + "";
+    // console.log("State is : ", filesize(objectSizeOf(state)));
```
7ruth: Some clean up here is needed

PR #726, opened 2017-10-27
Sticky Headers in Fabric List

shanberg:
Making headers in Fabric list sticky, to maintain context on scroll of long list.


shanberg:
This PR reveals an issue with the Breadcrumb element being 1px too tall, causing a 1px hole in in the header plane, but that's outside the scope of this fix, and should be resolved by the coming-soon new header.

PR #767, opened 2017-10-30
675 webpack modules dir

### LINKED ISSUE #675: Configure Webpack to have / resolve to project root and refactor external deps to absolute paths
Currently, / resolves to the filesystem, which forces us to use insane relative paths from deeply nested components. This looks awful and is super fragile. Moving to absolute paths off the project root would enable us to refactor components without breaking things.

spmcbride1201:
This PR adds the project's `src` directory as a modules directory. This allow us to resolve imports from `src` as pseudo-absolute paths in place of our existing insane relative paths. This will allow us to move files around without breaking all the import paths inside. Because there are so many relative paths, additional work cleaning this up will have to occur in follow-on PRs.

Resolves #675


spmcbride1201:
I think I need to add some WebPack tooling for the test pipeline to get this to work. A bit bigger scope that I had originally realized, so marking as `WIP:`

PR #768, opened 2017-10-30
New App Header ✨

### LINKED ISSUE #531: Add UI polish from Phase 1 mocks

### LINKED ISSUE #661: Implement Header Dashboard & Nav area from Phase 1 Mocks

### LINKED ISSUE #662: Implement App Bar

### LINKED ISSUE #663: Implement header/banner area from Phase 1 Mocks

### LINKED ISSUE #664: Implement new Tab component
```    - Tab
        Icon
        Title
        Detail
            Label
            Value
            Sparkline
        If tab is a container...
            Sub items extends Tab and stacks content vertically```

Consider responsive behavior, breakpoints.

### PR DESCRIPTION
**Pull request by kaitmore**
Resolves #664 
Resolves #662 
Resolves #663 
Resolves #661 
Resolves #531

### CODE REVIEW

**Comment by spmcbride1201** [comment]
Fabric Dashboard UX stuff:
- Feels pretty stable
- Summary Cards filter the Fabric dashboard as expected
- The sparklines inside of the summary cards seem frozen? Also not clear what they represent. I would think they they show the number of services, but that number likely wouldn't change very often. The charts appear to go up, so they likely are started at 0 and then go up to the actual total and stop rerendering
- Clicking a summary card clears the search filter, card/table toggle, and sort by / group by settings. I think we might want those to persist.
- Group By status and Sort By status probably don't make sense in the context of having "Down", "Warning", or "Stable" selected.
- The setting button looks a bit off to me on the dark background
- Responsiveness feels pretty good on Summary Cards

Service Dashboard UX:
- Feels a bit vertically off not having the App Header Summary Card nav elements. I wonder if it’s worth having a nav with a single card that’s always selected?

Instance Dashboard UX:
- Configuration doesn’t go anywhere. Since we only have Application level settings right now, perhaps we should just make Explorer a full-height card?
- The appearance of the scrollbar feels a bit jarring at times since it impacts the width of the App Header stuff.

**Comment by shanberg** [comment]
@spmcbride1201 Do you have any ideas about what else we can display in the details area of the Fabric tabs? The count of services is not very useful, to put it mildly.

File: src/components/AppContainer.js [hunk 1]
```diff
@@ -2,15 +2,12 @@ import styled from "styled-components";
 
 // toolbar wrapper - do not display scrollbar
 const AppContainer = styled.div`
-  -ms-overflow-style: none;
-  overflow-scrolling: auto;
   display: flex;
```
spmcbride1201: I see that you removed some of the vendor prefixes for IE. Did you test this there?

shanberg: I believe this was my change. Removed because we no longer need to override scrolling behavior because we don't have two separate scrolling contexts.

src/components/AppHeader/components/AppToolBar/index.js
```diff
@@ -0,0 +1,116 @@
...
+function AppToolBar({
+  pathname,
+  hideRoot,
+  appVersion = AppVersion,
+  toolbarButtons = defaultToolbarButtons
+}) {
+  return (
+    <AppToolBarHeader>
+      <BrandContainer>
+        <BrandLogo alt={`${BrandName} Logo`} src={decipherLogo} />
```
spmcbride1201: Is there any utility in having screen readers read this out? Maybe this should just be alt="" so that they skip it.

src/components/AppHeader/components/AppToolBar/index.js
```diff
@@ -0,0 +1,116 @@
...
+              pathname: "/",
+              search: ""
+            }}
+          >
+            {BrandName}
+          </Link>
+        </BrandText>
+      </BrandContainer>
+      <SkipNav
+        type="button"
```
spmcbride1201: It feels like this type and onKeyDown behavior is something that we'd probably want as our defaults. Can we shift this to styled-component attrs to accomplish this? Perhaps the props would be the ID of the thing to focus on, defaulting to main-content.

spmcbride1201: I guess this might be a moot point if we're expecting other apps to import the entire AppToolbar. My concern is just making sure the main API surface area of our components is easy to use via good defaults.

src/components/AppHeader/components/AppToolBar/index.js
```diff
@@ -0,0 +1,116 @@
...
+      <AppVersionLink
+        href="https://github.com/DecipherNow/gm-fabric-dashboard/blob/master/CHANGELOG.md"
+        rel="noopener noreferrer"
+        target="_blank"
+      >
+        {AppVersion}
+      </AppVersionLink>
+      <ButtonGroup toolbar>
+        {toolbarButtons.map(button => (
+          <NavButton
```
spmcbride1201: We probably need a way to style these buttons depending on the style of our AppToolbar.

src/components/AppHeader/components/Banner/components/Extra.js
```diff
@@ -0,0 +1,38 @@
...
+  BORDER_RADIUS_BASE,
+  COLOR_ALT_BACKGROUND
+} from "../../../../../style/styleVariables";
+import {
+  contrastColor,
+  spacingScale
+} from "../../../../../style/styleFunctions";
+
+const Extra = styled.a`
+  z-index: 2;
```
spmcbride1201: Can we add a comment here justifying the use of z-index? I suspect we're going to have to refactor this later and we're going to need to know what might break if we change this.

src/components/AppHeader/components/Banner/components/ExtrasLine.js
```diff
@@ -0,0 +1,5 @@
+import styled from "styled-components";
+
+const ExtrasLine = styled.div``;
```
spmcbride1201: Is this component necessary? This looks like it's just a div.

src/components/AppHeader/components/Banner/components/Header.js
```diff
@@ -0,0 +1,31 @@
+import styled from "styled-components";
+import {
+  FONT_SIZE_HERO,
+  COLOR_ALT_BACKGROUND
+} from "../../../../../style/styleVariables";
+
+const Header = styled.h1`
```
spmcbride1201: If these z-indexes are just used internally for the pseudo-element, we should probably accept a base z-index prop and then internally calculate these values.

src/components/AppHeader/components/Banner/components/ExtrasLine.js
shanberg: From a layout perspective, we do need a wrapping element there. But you're right, it doesn't need to be a styled component.

src/components/AppHeader/components/InstanceHeaderContent/index.js
```diff
@@ -0,0 +1,102 @@
+// import _ from "lodash";
```
spmcbride1201: zombies

src/components/AppHeader/components/InstanceHeaderContent/index.js
```diff
@@ -0,0 +1,102 @@
...
+import { PropTypes } from "prop-types";
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import { withRouter } from "react-router-dom";
+
+// import SidebarCard from "./SidebarCard";
```
spmcbride1201: zombies

src/components/AppHeader/components/InstanceHeaderContent/index.js
```diff
@@ -0,0 +1,102 @@
...
+    const {
+      basePath,
+      metrics,
+      runtime,
+      // serviceName,
+      // services,
+      // instanceID,
+      headerTabs
+    } = this.props;
+    // Render Services if the path does not contain a selected service or instances
```
spmcbride1201: zombies

src/components/AppHeader/components/InstanceHeaderContent/index.js
```diff
@@ -0,0 +1,102 @@
...
+    //       key={instanceID}
+    //       title={instanceID}
+    //     />
+    //   );
+    // });
+
+    switch (runtime) {
+      case "JVM":
+        return (
+          <JVMSidebarContent
```
spmcbride1201: We should rename Sidebar to something more semantically neutral probably.

src/components/AppHeader/components/InstanceHeaderContent/index.old.js
```diff
@@ -0,0 +1,195 @@
+import _ from "lodash";
+import { PropTypes } from "prop-types";
```
spmcbride1201: Why keep this file if it's all zombie content?

src/components/AppHeader/index.js
```diff
@@ -1,6 +1,57 @@
...
+  // This will display the first item in the path
+  // Else if we are on an instance, it will display serviceName : last 8 digits of instanceID
+  // If the path is on the root, display Fabric
+  const path = pathname
+    .replace(/^\/|\/$/g, "")
+    .replace("%2F", "/") // String out escaped slashes if found
+    .split("/");
+
+  // If the selected Instance matches what's in our path, then display it, else return an empty string
+  let instance = selectedInstance === path[2] ? ` : ${path[2].substr(-8)}` : "";
```
spmcbride1201: This should probably use the trimID util function to make sure that we keep all of the logic for how we truncate IDs in a single place.

spmcbride1201: What are the circumstances where the selectedInstance doesn't match the URL path? I can't recall how we're setting the Redux state off the top of my head.

src/components/AppHeader/scenes/NoSDS/index.js
```diff
@@ -0,0 +1,24 @@
...
+  runtime: PropTypes.string
+};
+
+/**
+ * Main Sidebar component
+ * @export
+ * @returns JSX.Element
+ */
+function Header({ runtime }) {
+  return <InstanceHeaderContent runtime={runtime} />;
```
spmcbride1201: We'll have to run the dashboard without the SDS to test this.

src/components/AppHeader/scenes/UseSDS/components/AppHeaderContent/index.js
```diff
@@ -0,0 +1,25 @@
...
+import InstanceHeaderContent from "../../../../components/InstanceHeaderContent";
+
+// Add a route for /:serviceName/:serviceVersion
+
+function AppHeaderContent() {
+  return (
+    <Switch>
+      {/* Match on root and stable, down, or warning */}
+      <Route
+        path="/:path(|stable|down|warning)"
```
spmcbride1201: This is nifty syntax!

src/components/AppHeader/scenes/UseSDS/components/AppHeaderContent/scenes/Fabric/index.js
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <TabNav>
+      <Tab
+        title="All Services"
+        href="/"
+        icon="grid"
+        lines={[
+          {
+            name: "Services",
+            value: statusCount.total
```
spmcbride1201: Can you open an Issue flagging the task to replace this with something dynamic and add a TODO comment referencing the Issue number?

src/components/AppHeader/scenes/UseSDS/components/AppHeaderContent/scenes/Fabric/old.js
```diff
@@ -0,0 +1,45 @@
+import _ from "lodash";
```
spmcbride1201: This is all zombie content, right? Can we delete?

src/components/AppHeader/scenes/UseSDS/index.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import FabricAppHeaderContent from "./components/AppHeaderContent";
+
+/**
+ * Main Sidebar component
+ * @export
+ * @returns JSX.Element
+ */
+export default function Header() {
+  return <FabricAppHeaderContent />;
```
spmcbride1201: Looks like AppHeaderContent should be refactored to be the UseSDS scene.

src/components/Main/scenes/Fabric/FabricGrid.js
```diff
@@ -207,8 +206,4 @@ class FabricGrid extends Component {
   }
 }
 
-function mapStateToProps({ fabric: { services } }) {
-  return { services: _.values(services) };
-}
-
-export default connect(mapStateToProps)(FabricGrid);
+export default withRouter(FabricGrid);
```
spmcbride1201: Why was withRouter added here? Were children not rendering as expected?

src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/index.js
```diff
@@ -64,11 +64,15 @@ export default function SectionCardsView({
...
     const dataGroupedByHeader = _.groupBy(services, item => item.headerTitle);
-    // If we are grouping by state, we always want to group our services in the order "Down, Warning, Stable"
+    const headerTitles = Object.keys(dataGroupedByHeader);
+
+    // If the groupByAttribute is set to "Status" and we have 3 header titles, that means we are looking at the whole
+    // Fabric Grid, so we need to render headers in the order "Down, Warning, Stable". Otherwise we can just render the headers normally.
+
     const headers =
-      groupByAttribute === "Status"
+      headerTitles.length === 3 && groupByAttribute === "Status"
```
spmcbride1201: This is a weird magic number situation. Can you look at the URL to determine this instead?

src/index.js
```diff
@@ -32,15 +30,13 @@ UIkit.use(Icons);
...
-
 ReactDOM.render(
   <Provider store={store}>
     <Router history={history}>
       <AppContainer>
-        <Sidebar fabricServer={fabricServer} />
-        <Main className="uk-width-5-6@s" fabricServer={fabricServer} />
+        <Header />
+        {/* <Sidebar /> */}
+        <Main className="uk-width-5-6@s" />
```
spmcbride1201: It seems like Main shouldn't need this layout class anymore.

src/utils/selectors.js
```diff
@@ -188,8 +187,41 @@ export const getSideBarContent = createSelector(getServices, services =>
...
+    status = computeStatus(
+      service.instances.length,
+      service.minimum,
+      service.maximum
+    );
+    if (status === "Stable") {
+      ++stableServicesCount;
+    } else if (status === "Warning") {
+      ++warningServicesCount;
+    } else {
```
spmcbride1201: This should still be an explicit else if in case computeStatus doesn't work as expected.

src/utils/selectors.js
```diff
@@ -188,8 +187,41 @@ export const getSideBarContent = createSelector(getServices, services =>
...
+ * returns an object with a count for each status
+ * @param {Object[]}
+ * @returns {Object}
+ */
+export const getStatusCount = createSelector(getServices, services => {
+  let stableServicesCount = 0,
+    warningServicesCount = 0,
+    downServicesCount = 0,
+    status;
+  _.values(services).forEach(service => {
```
spmcbride1201: I think the lodash function you're looking for is `_.countBy`.

stories/AppHeader.js
```diff
@@ -86,32 +89,98 @@ storiesOf("App Header", module)
...
+                />
+              );
+            })}
+          </TabGroup>
+        </TabNav>
+      </AppHeaderContainer>
+    );
+  });
+
+// <AppHeader
```
spmcbride1201: zombies

spmcbride1201:
IE11 Notes:
- The Warning services don't wrap as expected, causing the width of the app to be like 300% of the width of the browser window.
- The Footer is not at the bottom of the app. It seems like some z-index hackery is causing it to be behind a chart
- Threads Buttons are squished
- Reusable text cards (blanking on the name) are messed up

spmcbride1201:
No regressions noticed in Safari or Firefox ESR.

src/components/AppHeader/components/Banner/components/Header.js
shanberg: Setting any z-index on any parent element creates a new relative stacking context for its children, so z-index isn't an absolute value that needs calculating.

spmcbride1201: Ok, cool. I hadn't realized that. What if we were going to reuse this component outside of this app? It seems like if our components are using z-index to achieve something, the highest ancestor with a z-index would still need to be set via props in case other sibling element in the app are using z-index differently.

src/components/Main/scenes/Fabric/FabricGrid.js
kaitmore: @spmcbride1201 
So in the router, I changed 
`<Route exact path="/" component={FabricGrid} />` 
to 
`<Route
        exact
        path="/"
        render={() => <FabricGrid services={_.values(services)} />}
      />`
Would it be better to just pass `location` to that render function instead of using `withRouter`?

src/components/AppHeader/index.js
kaitmore: @spmcbride1201 When you have a selectedInstance and then navigate away, it doesn't reset to an empty string. So after you navigate to an instance it would always display in the header.

But anyways, I know this sucks. I'm going to open an issue to refactor.

src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/index.js
kaitmore: @spmcbride1201 Refactored to grab header from URL.

src/components/AppHeader/components/AppToolBar/index.js
kaitmore: @spmcbride1201 Done

src/components/AppHeader/components/AppToolBar/index.js
kaitmore: @spmcbride1201 I like this idea. I refactored it with a nifty attribute in the styled component. You can now just pass it the ID of a DOM element where you want it to jump.

src/components/AppHeader/scenes/UseSDS/components/AppHeaderContent/scenes/Fabric/index.js
kaitmore: @spmcbride1201 Done #786

src/utils/selectors.js
kaitmore: Issue #788

src/components/AppHeader/components/InstanceHeaderContent/index.js
kaitmore: Done

src/components/AppHeader/scenes/UseSDS/index.js
kaitmore: @spmcbride1201 Done

PR #772, opened 2017-10-31
Fix: Cursor for search input clear button

### LINKED ISSUE #755: Pointer should be cursor when hovering over the button to clear the search input.
The cursor currently remains `text` when hovering over the clear button. This should be changed to be a `cursor`
image

kaitmore:
Resolves #755


spmcbride1201:
Currently fixed on Chrome and Safari, since they are both part of the webkit family. Firefox doesn't seem to offer a clear button for their search input. IE11 has one, but it's not styled.

spmcbride1201:
The IE11 vendor prefix seems to be this: https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-clear

kaitmore:
@spmcbride1201 I'm stuck on this. I'm able to use `::-ms-clear` to apply styles to the clear button, except for the cursor. I haven't found any discussions online about it either. IE11 should accept `cursor: pointer`, but I also tried `cursor: hand`, still no luck.

spmcbride1201:
@kaitmore. Ok, noble effort anyways. I'll accept and you can merge as is.

PR #776, opened 2017-10-31
chore: Wipe out submodule

### LINKED ISSUE #678: Fatal error when updating submodules in the git repo
An error stating "fatal: No url found for submodule path 'docker/docker-graphite-statsd' in .gitmodules" is thrown when the following command is run (note: this command is automatically run in Sourcetree when performing a pull operation).

Command: `git -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree submodule update --init`

spmcbride1201:
Something seems to have gotten messed up a while ago, but the zombie submodule references have been resolved. 

Resolves #678


jsweetland:
The submodule update command that caused problems previously now succeeds, LGTM!

Issue #745, opened 2017-10-30
Routes Table Line Item Error % Issues


spmcbride1201:
1. The error % is resolving to NaN in some cases
2. The error % is using a yellow color that is very hard to make out on a white background.

Note: this was seen in a JVM microservice, but it likely is also occurring with other runtimes.

![image](https://user-images.githubusercontent.com/8700322/32186379-501507b0-bd78-11e7-9bf7-fc867d75500e.png)

karinakeen:
@spmcbride1201 @shanberg Which colors do you want against white/black background for stable/warning/error?

spmcbride1201:
I defer to @shanberg on this.

karinakeen:
blocked by #782

PR #778, opened 2017-10-31
743 sort alpahbetically by name by default

### LINKED ISSUE #743: Sorting Services by Status should also sort alphabetically by name within each status.

### LINKED ISSUE #746: When sorting by Error % or Latency, the default sort should be high to low
Currently, when a user selects sort by Error % for our tables, the top result is the lowest error rate. This order should be flipped. This should be the same for latency metrics as well.

### LINKED ISSUE #775: Go dashboards should have Sort By Error % instead of Sort By Error Count

karinakeen:
This PR:

Fix: Sorting issues

- Sorting Services by Status sort alphabetically by name within each status
- When sorting by error % or latency, default sorting is in descending order
- Display Sort By Error % instead of Sort By Count

Resolve #743, resolve #746 and resolve #775


karinakeen:
@spmcbride1201 Looks good, this rewrites my fix in PR #745 😄

Issue #763, opened 2017-10-30
Route name clipped prematurely at 200% zoom


spmcbride1201:
![image](https://user-images.githubusercontent.com/8700322/32191745-78bc9646-bd88-11e7-8c3e-83abc03708ef.png)

PR #784, opened 2017-10-31
#763 Route name clipped prematurely at 200% zoom

7ruth:


src/components/Main/scenes/Instance/scenes/go/scenes/Routes/components/RoutesTable/components/LineItem.js
```diff
@@ -83,7 +83,7 @@ export default class RoutesTableLineItem extends Component {
         role="link"
         tabIndex="0"
       >
-        <TableCol vizBar>
+        <TableCol vizBar customFlex="1 1 20%">
```
kaitmore: It looks like the header is off from the column now. I think you can just add the same customFlex to the corresponding header in Routes/index.js
screen shot 2017-11-01 at 12 40 11 pm

7ruth: Fixed! Great catch!

spmcbride1201:
@kaitmore - please check out the changes and approve if ready.

PR #798, opened 2017-11-01
Fix: Explorer layout bug

kaitmore:


src/components/Main/components/AppContent.js
```diff
@@ -4,6 +4,10 @@ import { COLOR_CONTENT_BACKGROUND } from "style/styleVariables";
 // content wrapper - display scrollbar
 const AppContent = styled.div`
   background-color: ${COLOR_CONTENT_BACKGROUND.string()};
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
```
shanberg: `min-height` of 100vh is causing layout overflow because the AppContent sits below the AppHeader, so the total layout height ends up being 100vh + whatever height the AppHeader is.

shanberg:
Looks like the `AppContainer` element also needs to set a flexbox context for its children, so that the flex properties in `AppContent` will apply.

PR #802, opened 2017-11-01
feat: Cleaned up Explorer style to better match app

shanberg:
- Removed background color
- Rewrote layout properties
- Minor adjustments to element spacing


shanberg:
I'm not clear on why this broke in testing. @spmcbride1201 can you weigh in?

kaitmore:
@shanberg I think you have some unused variables

Issue #792, opened 2017-11-01
Center main content body


kaitmore:
<img width="520" alt="screen shot 2017-11-01 at 10 24 11 am" src="https://user-images.githubusercontent.com/22768918/32279336-d8a9839a-beee-11e7-8278-2ae474cb69b0.png">

shanberg:
- [ ] Proper width on Toolbars
- [ ] Proper width on Fabric Grid
- [ ] Proper width on All Tables
- [ ] Proper width on Service Summary
- [ ] Proper width on Grid pages, charts and tables
- [ ] Proper width on Explorer

shanberg:
Closed by #809

PR #809, opened 2017-11-01
Center content with reasonable max-width

shanberg:
- Page contents (everything below the app header) now constrained to 1300px wide and centered.

Addresses #792


kaitmore:
@shanberg Love how this looks! Small thing - I noticed that the table view content is centered but the card view content isn't.

shanberg:
The solution to that might be pretty complicated. I’ll have to rewrite the card sizing. Would you mind filing a new issue for that?

PR #819, opened 2017-11-02
785 fix storybook

### LINKED ISSUE #677: Refactor Storybook stories into the component file structure
The intention here is that the stories for each component should reside inside of the component's directory, rather than in a top-level `/stories` directory. For example, stories for `AppToolbar` would reside in a file at `/src/components/Main/components/AppToolbar/index.stories.js`.

If the storybook was just targeting a single styled-component, there wouldn't be a directory, so the stories would reside in the same `components` directory as the styled-component with the same name and a `.stories.js` suffix. E.g. `awesome.js` would be accompanied by `awesome.stories.js`.

Instructions for configuring this are at https://storybook.js.org/basics/writing-stories/#loading-stories-dynamically

### LINKED ISSUE #785: storybook is broken

spmcbride1201:
Storybook now uses our dev webpack config and is configured to allow our stories to be loaded out of our `src` directory to allow the stories to live next to the components they're covering.

Resolve #785 
Resolve #677


.storybook/config.js
```diff
@@ -2,8 +2,11 @@ import { configure } from "@storybook/react";
 import { setDefaults } from "@storybook/addon-info";
 import { setOptions } from "@storybook/addon-options";
 
+const req = require.context("../src", true, /\.stories\.js$/);
+
 function loadStories() {
-  require("../stories");
+  req.keys().forEach(filename => req(filename));
```
kaitmore: This is really cool!!

PR #826, opened 2017-11-02
715 match threads value in app header and content

### LINKED ISSUE #715: Threads value in the sidebar does not match value on page
The Threads value shown in the sidebar does not match the Threads value shown on the main page. Discovered in the f5nqgzxp4e8000000000 instance of Network Internet Information service, see screenshot.
screen shot 2017-10-26 at 1 38 09 pm

karinakeen:
This PR resolves #715 

refactor: JVM app header fetches thread count from selector


src/components/AppHeader/components/InstanceHeaderContent/scenes/JVM.js
```diff
@@ -42,7 +48,7 @@ export default function SidebarContent({ basePath, metrics, headerTabs }) {
         lines={[
           {
             name: "Threads",
-            value: getLatestAttribute(metrics, "jvm/thread/count")
+            value: threadCounts.all
```
tilleryd: Since we're already passing in the `metrics` data, the data is already there and we should leverage that rather than creating a separate prop just for the threads.  Instead of passing a prop which increases the number of dependencies between components you might be able to do something similar to the `UpTime` component implementation in https://github.com/DecipherNow/gm-fabric-dashboard/pull/837.

karinakeen: @spmcbride1201 
I created a Threadcounts library component that returns threadCounts object.  This is resulting in a similar bug in `Uptime`.  'ThreadCounts` is not updating after the initial fetch in app header, while the props are updating in content.

screen shot 2017-11-03 at 12 22 11 pm

screen shot 2017-11-03 at 12 22 07 pm

karinakeen:
@spmcbride1201 Changes implemented in JVM service in files

`src/components/AppHeader/components/InstanceHeaderContent/scenes/JVM.js` and 

`src/components/library/ThreadCounts.js`

tilleryd:
LGTM

@spmcbride1201 This is good to be merged.

PR #837, opened 2017-11-02
714 fix: match uptime value in app header

### LINKED ISSUE #714: Uptime value in the sidebar does not match value on page
The uptime value shown in the sidebar does not match the uptime value shown on the main page. Discovered in the f5nqgzxp4e8000000000 instance of Network Internet Information service, see screenshot.
screen shot 2017-10-26 at 1 18 05 pm

tilleryd:
This PR resolves #714 ...first attempt.

** Check for potential performance issue with store updating every second **  

Fix:  Uptime in header now matches uptime in content

Feat: Content updates redux store with uptime value

Refactor: App header gets uptime value from store


tilleryd:
@karinakeen I created a new `UpTime` component that can be re-used throughout the app when a service up time needs to be displayed.  This resolves the performance hit of updating the store every 1 sec by taking the service `start_time` and updating the up time every second.  I copied the code from `gm-fabric-dashboard/src/components/AppHeader/components/InstanceHeaderContent/scenes/JVM.js`; specifically the logic that manages the 1 second interval to update the up time.  Now this logic lives in one place in `gm-fabric-dashboard/src/components/library/UpTime.js`.

I used a new pattern called a render prop in `UpTime`.  It's basically a method to dynamically compose React components.  A render prop is defined as "a function prop that a component uses to know what to render" and you can see the evolution of component composition in this video - https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce.

TODO:
- please remove the code related to `selectedInstanceUptime`
- please apply `UpTime` to the Go service similar to how I implemented it for the JVM service

src/components/Main/components/ReadoutItem/index.js
```diff
@@ -61,5 +61,9 @@ ReadoutItem.propTypes = {
   iconName: PropTypes.string,
   iconShape: PropTypes.string,
   title: PropTypes.string,
-  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+  value: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+    PropTypes.object
```
tilleryd: @spmcbride1201 I had to expand the `value` propTypes to also take a React component to support the usage of the new `UpTime` component.

karinakeen:
I applied Uptime to Go service, but it is displaying some odd behavior of syncing on refresh only. 
  Will investigate more later.

src/jumpstate/settings.js
```diff
@@ -32,10 +32,10 @@ const settings = State({
 
     /**
      * These attributes contain static settings that the dashboard uses to directly
-     * poll an inidividual microservice when running without a "Fabric Server"
+     * poll an individual microservice when running without a "Fabric Server"
      * discovery service. They are populated when the app initializes from
      * meta tags in the index.html. When the dashboard is used with a "fabric server"
-     * (discovery service), this value is always null, as the disovery service
+     * (discovery service), this value is always null, as the discovery service
```
tilleryd: @karinakeen Good catches on the mis-spells.

tilleryd:
@karinakeen I fixed the issue where the up time was not being updated in the App Header.  I added the `componentWillReceiveProps` lifecycle function to `UpTime` to set internal state when the `startTime` prop value changes.

src/components/Main/scenes/Instance/scenes/go/scenes/Summary.js
```diff
@@ -28,17 +30,9 @@ class SummaryGrid extends Component {
   };
 
   state = {
-    _timer: null,
-    startTime: getLatestAttribute(this.props.metrics, "system/start_time"),
-    uptime: 0
+    startTime: getLatestAttribute(this.props.metrics, "system/start_time")
```
spmcbride1201: why are we forking state here?

src/components/Main/scenes/Instance/scenes/jvm/scenes/Summary.js
```diff
@@ -24,17 +24,9 @@ class SummaryGrid extends Component {
   };
 
   state = {
-    _timer: null,
-    startTime: getLatestAttribute(this.props.metrics, "jvm/start_time"),
-    uptime: 0
+    startTime: getLatestAttribute(this.props.metrics, "jvm/start_time")
```
spmcbride1201: ditto. Why are we forking state here?

src/components/library/UpTime.js
```diff
@@ -0,0 +1,48 @@
...
+    const uptime =
+      this.state.startTime > 0 ? Date.now() - this.state.startTime : 0;
+
+    this.setState({
+      uptime: convertMS(uptime)
+    });
+  }
+
+  render() {
+    return <div>{this.props.render(this.state)}</div>;
```
spmcbride1201: Can we let the render func decide the wrapping element? This would make this component way more flexible.

src/components/Main/scenes/Instance/scenes/go/scenes/Summary.js
tilleryd: I just refactored the code that was already there.  Is the comment above `componentWillReceiveProps` no longer valid?

```
  // handles edge case when start_time changes
  componentWillReceiveProps(nextProps) {
...
```

tilleryd: Actually... 🤔

tilleryd: @spmcbride1201 I refactored Summary components for JVM and GO to be stateless functions.

tilleryd:
@spmcbride1201 I'm pretty happy with how this turned out now.  We moved all the logic to determine service up time to the new `UpTime` component and as a result the JVM and GO Summary components are completely stateless now.  #winning

PR #844, opened 2017-11-03
Fix: Disable links to microservices with no metrics

### LINKED ISSUE #721: We are allowing users to open instance dashboards for microservices that don't have metrics
This causes the user to open a dashboard that immediately begins erroring out. If you have access to edge, you can see this at https://edge.deciphernow.com/#/Odrive%20Service/1.0/a2385c213e62416636f524c8c1eecbc3/summary

We should treat `metered` in the same places as `authorized` with the logic that the dashboard must be `metered & authorized` for a link to be present on the fabric dashboard.

If the user directly hit the service via a URL, the user should be redirected back to the fabric.

[
...
  {
    "name": "Odrive Service",
    "version": "1.0",
    "owner": "MEME",
    "capability": "Utility",
    "runtime": "GO",
    "documentation": "https://edge.deciphernow.com/services/object-drive/1.0/apps/drive/index.html",
    "minimum": 1,
    "maximum": 1,
    "authorized": true,
    "metered": false,
    "threaded": false,
    "instances": [
        "a2385c213e62416636f524c8c1eecbc3"
    ]
  },
...
]

kaitmore:
- Added logic to Fabric router that checks if a service is both authorized & metered
- Cleaned up the card/list item code by using an onClick handler to [disable react-router Links](https://github.com/ReactTraining/react-router/issues/1082#issuecomment-160348163)
- Added a notification if a user attempts to navigate to an unmetered service

Resolves #721


src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/components/GMServiceCardCollection/components/GMServiceCard/index.js
```diff
@@ -93,28 +93,27 @@ export default function GMServiceCard({
...
-          </Title>
-        </ServiceLink>
-      ) : (
-        <ServiceInfo>
-          <Title title={titleNameAttribute} cardFontWeight={cardFontWeight}>
-            {titleName}
-          </Title>{" "}
-        </ServiceInfo>
-      )}
+      <ServiceLink
```
spmcbride1201: Looks much cleaner!!

src/components/Main/scenes/Fabric/FabricGrid.js
```diff
@@ -47,10 +47,18 @@ class FabricGrid extends Component {
         timeout: 0
       });
       // Reset location state
-      const location = {
+      history.replace({
```
spmcbride1201: Good job getting rid of intermediate variable!

src/components/Main/scenes/Fabric/components/FabricMainView/scene/Table/components/GMServiceList/components/GMServiceListItem/index.js
```diff
@@ -35,33 +35,29 @@ export default class GMServiceListItem extends Component {
...
+              cursor={
+                status !== "Down" && authorized && metered
+                  ? "pointer"
+                  : "not-allowed"
+              }
+              tabIndex="0"
+              disabled={status === "Down"}
+            >
+              <IconWrapper>
+                {groupByAttribute.toLowerCase() !== "status" && (
```
spmcbride1201: Why not have this wrap IconWrapper? It seems like an IconWrapper without an Icon is a pointless DOM element (unless removing it messes up the layout).

kaitmore: It does mess up the layout. Perhaps this should be refactored at some point though.

PR #857, opened 2017-11-05
WIP: Added PurifyCSS plugin

kaitmore:
Implemented the purifycss plugin to reduce our css bundle size. This plugin is also handling minification now, rather than css-loader.


spmcbride1201:
This seems to be purifying our CSS a bit too much. Pretty much all of our scss-generated stuff seems to be getting zapped. I'll fiddle with the config a bit to see if I can get it to work.

kaitmore:
@spmcbride1201 I think I got it to include our scss with this latest commit, however I'm still unclear on how to handle styled-components.

spmcbride1201:
I think I'm going insane trying to get this configured right. It looks like no matter how I set this, the loader crawls through the node_modules directory, causing us to retail tons of unused UIKit styles. I think the behavior we want is to scan our post-Babel-post-Webpack JS bundle for class names and only write those classnames to the CSS bundle. The API seems pretty simple, but I couldn't achieve this after spending a few hours on this. Unless @tilleryd wants to also take a whack at this, I suggest we give up on this effort.

spmcbride1201:
Here is the last set of configurations I was beating my head against:
```js
    new PurifyCSSPlugin({
      paths: glob.sync(path.resolve(paths.appBuild)),
      verbose: true,
      purifyOptions: {
        info: true,
        minify: true
      }
    })
```

spmcbride1201:
I spoke with Tillery, and we think we should abandon the effort. Thank you for working on this!

PR #861, opened 2017-11-05
582 escape url

### LINKED ISSUE #582: Escape Spaces in URL
Currently, the URL includes parameters that have actual spaces, such as the following:
```
https://edge.deciphernow.com/#/Discovery Service/1.0/1ea349346c473b5c50ef9d8899d73229/route
```
When URLs like this are pasted into slack and other chat clients that dynamically link the URL, they interpret the space as terminating the URL. We should use encodeURI() on all parameters to fix this.

tilleryd:
Fixes #582 - spaces in service names in the URL.  Since we're placing labels in sentence structure into the URLs we basically needed to do this on our end.  But the decoding of URLs was unexpected and I tracked down a known feature or bug or both of `history`.  

`react-router` uses `history` which auto decodes location objects.  The `history` authors say that it was done that way to return whatever the user types in the URL bar.  Spaces and all; not an encoded URI.  Thus, if you use `encodeURI()` react-router will return you a decoded string e.g. "%20" will get converted to " " (empty space).

Issue:
https://github.com/ReactTraining/history/issues/516

PR TODO:

- [x] Fix clicking on breadcrumb link
- [x] Unit tests


src/utils/index.js
```diff
@@ -99,3 +99,38 @@ export function formatAsDecimalString(number, numberOfDecimals = 3) {
     minimumFractionDigits: numberOfDecimals
   });
 }
+
+/**
+ * Utility function to generate a URL safe string by replacing spaces with underscores
+ *
+ * TECH DEBT:
```
tilleryd: @spmcbride1201 I came at this from an investigatory process to resolve the mystery of the decoding URI.  That's why this comment is here.  I feel like it's good to capture this limitation but perhaps this isn't the right place for it...

src/utils/index.js
```diff
@@ -88,3 +88,38 @@ export function formatAsDecimalString(number, numberOfDecimals = 3) {
...
+ * Refer to encodeParameter
+ *
+ * @export
+ * @param {string} param
+ * @returns string
+ */
+export function decodeParameter(param) {
+  return param.replace(/_/gi, " ");
```
tilleryd: TODO: This will need to take into account service names that already have _ in them.

src/components/AppHeader/components/AppToolBar/index.js
```diff
@@ -72,6 +73,7 @@ function AppToolBar({
           .replace("%2F", "/") // String out escaped slashes if found
           .split("/")
           .map((val, idx) => {
+            val = decodeParameter(val);
```
spmcbride1201: Not necessarily a problem because we aren't using _s, but this attempts to decode all parts of a Breadcrumb.

PR #884, opened 2017-11-06
fix vizbar styled component position

### LINKED ISSUE #874: Mysterious Blue Bar just above footer when viewing Routes page for JVM
Only appears on this view. Does not seem to have anything to do with focus.

image

karinakeen:
Resolve #874


karinakeen:
@shanberg @kaitmore review pls

PR #895, opened 2017-11-07
Increase Net Change Calculations to three decimal points

### LINKED ISSUE #875: Requests Per Second Calculation
Currently, the calculation of Request Per Second seems to produce an integer. This results in the situation where if a service receives 4 HTTP requests over five seconds, the dashboard reports 0 requests/s. This causes the graphs to be imprecise at low rates to change. To fix this, the requests per second calculation should be calculated and graphed out to three decimal points. This would show greater activity in the sparklines and charts under low rates of change.

spmcbride1201:
I modified mock-sds so that the https/requests metrics for JVM microservices randomly increment. You should be able to see the more precise net change behavior on the Summary tab of any JVM microservice.

Resolves #875


7ruth:
screen shot 2017-11-07 at 3 55 59 pm
That looks awesome ^

PR #902, opened 2017-11-07
Adding propsTypes to styled components

### LINKED ISSUE #727: Add PropTypes for styled-components
As our styled-components get more complex, it might be beneficial to add PropType validation. In addition to catching bugs I think it would be a nice form of documentation.  Since everything is broken out into individual files now this makes more sense.

7ruth:
Resolves #727


src/components/Main/components/Readout/components/ReadoutItem/components/ReadoutItemData.js
```diff
@@ -15,4 +16,8 @@ const ReadoutItemData = styled.div`
   }
 `;
 
+ReadoutItemData.propTypes = {
+  paddingLeft: PropTypes.string
```
spmcbride1201: could be a number here I think

src/components/Main/components/TableCol.js
```diff
@@ -10,4 +11,8 @@ const TableCol = styled.div`
     props.errorPercent ? `color: ${errorColor(props.errorPercent)}` : ""};
 `;
 
+TableCol.propTypes = {
+  errorPercent: PropTypes.string
```
spmcbride1201: could be a number

src/components/Main/components/Readout/components/ReadoutItem/components/ReadoutItemData.js
7ruth: I checked our code base for this and I actually didn't see anywhere where paddingLeft prop is passed to readoutItemData...so I went with what I thought was logical. Adjusting to include both.

src/components/Main/components/TableCol.js
7ruth: I checked 3 files where TableCol gets this prop (FunctionTable/../LineItem.js; RoutesTable/../LineItem.js; RoutesTableLineItem.js) and in all of those file the prop is a string, so unless you think that will change in the future I'd want to keep this as it is.

karinakeen: @7ruth They could be number, depending on how they get refactored for rendering..You could use PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.number
  ])

7ruth: Ok sounds like you expect it to change in the future... will adjust.

PR #906, opened 2017-11-08
Fix: Sort Select Dropdown flickering on route change

### LINKED ISSUE #853: Sort Select flickers in and out when clicking different summary cards on Fabric Dashboard

kaitmore:
Resolves #853  

I moved the `statusView` state from FabricGrid to the top level router in `generateStatusRoutes.js`. `statusView` will default to false for the root route.


src/components/Main/scenes/Fabric/FabricGrid.js
```diff
@@ -185,7 +180,7 @@ class FabricGrid extends Component {
   };
 
   render() {
-    const { services = [] } = this.props;
+    const { services = [], statusView = false } = this.props;
```
tilleryd: I think the default value of statusView should be set using React's `defaultProp` pattern.

```
  ...
  static defaultProps = {
    statusView: false
  };
  ...
  render() {
    const { services = [] } = this.props;
    ...
  }
```

kaitmore:
@tilleryd Implemented defaultProps in FabricGrid

PR #908, opened 2017-11-08
Test: Fabric Grid

kaitmore:
A first go at testing this pretty sizable component. Be prepared to see lots of console warnings when you run `npm test`. We had to work around the jumpstate calls in FabricGrid's lifecycle methods.

As per discussion, we will come back to this and complete more end-to-end testing at a later date.

Also, this PR includes a tiny refactor of FabricGrid. It's no longer wrapped in withRouter, instead it's passed props in the render method of the Route.


src/components/Main/scenes/Fabric/FabricGrid.test.js
```diff
@@ -0,0 +1,284 @@
+import React from "react";
+import { shallow, mount } from "enzyme";
+import FabricGrid from "./FabricGrid";
+import { MemoryRouter, Route } from "react-router";
+import _ from "lodash";
+import "store/states/fabric.js";
```
tilleryd: I believe our standard is to separate 3rd party and source code import statements.

src/components/Main/scenes/Fabric/FabricGrid.test.js
```diff
@@ -0,0 +1,284 @@
...
+
+describe("Fabric Grid Main View", () => {
+  beforeEach(() => {
+    FabricGridWrap = mount(RouterWrap(["/"]));
+  });
+
+  test("renders all services in card view", () => {
+    expect(FabricGridWrap.find(FabricGrid).find("GMServiceCard").length).toBe(
+      3
+    );
```
tilleryd: Perhaps there should be a negative test case to confirm there are no rendered `GMServiceListItem ` components.

src/components/Main/scenes/Fabric/FabricGrid.test.js
```diff
@@ -0,0 +1,284 @@
...
+
+  test("renders all services in table view", () => {
+    // Remount with new search query
+    FabricGridWrap = mount(RouterWrap([{ search: "?viewType=Table" }]));
+
+    let FabricGridInstance = FabricGridWrap.find(FabricGrid).instance();
+
+    expect(
+      FabricGridWrap.find(FabricGrid).find("GMServiceListItem").length
+    ).toBe(3);
```
tilleryd: Perhaps there should be a negative test case to confirm there are no rendered `GMServiceCard` components.

src/components/Main/scenes/Fabric/FabricGrid.test.js
```diff
@@ -0,0 +1,284 @@
...
+    expect(FabricGridInstance.state.groupByAttribute).toBe("Capability");
+
+    FabricGridInstance = mount(RouterWrap([{ search: "?groupBy=Status" }]))
+      .find(FabricGrid)
+      .instance();
+
+    FabricGridInstance = mount(RouterWrap([{ search: "?groupBy=None" }]))
+      .find(FabricGrid)
+      .instance();
+    expect(FabricGridInstance.state.groupByAttribute).toBe("None");
```
tilleryd: I think for each of these tests we should test that services are properly grouped in the outputted HTML.

src/components/Main/scenes/Fabric/FabricGrid.test.js
```diff
@@ -0,0 +1,284 @@
...
+      .find(FabricGrid)
+      .instance();
+
+    expect(FabricGridInstance.state.sortByAttribute).toBe("Name");
+
+    FabricGridInstance = mount(RouterWrap([{ search: "?sortBy=Status" }]))
+      .find(FabricGrid)
+      .instance();
+
+    expect(FabricGridInstance.state.sortByAttribute).toBe("Status");
```
tilleryd: I think for each of these tests we should test that services are properly sorted in the outputted HTML.

src/components/Main/scenes/Fabric/FabricGrid.test.js
```diff
@@ -0,0 +1,284 @@
...
+    );
+    expect(FabricGridWrap.html().includes("Entry Monitoring")).toBe(false);
+    expect(
+      FabricGridWrap.html().includes("Grace Hopper Battleship Control Program")
+    ).toBe(true);
+  });
+
+  test("has buttons that toggle card and table view", () => {
+    let FabricGridInstance = FabricGridWrap.find(FabricGrid).instance();
+    const button = FabricGridWrap.find(FabricGrid).find("button");
```
tilleryd: If more buttons are added to the toolbar this test will break.

src/components/Main/scenes/Fabric/FabricGrid.test.js
```diff
@@ -0,0 +1,284 @@
...
+      //Wait another 500ms
+      await setTimeout(() => {
+        expect(FabricGridInstance.state.lastPushedQueryString).toBe(
+          "groupBy=Owner"
+        );
+        expect(FabricGridInstance.state.lastPushedQueryString).not.toBe(
+          "viewType=Table"
+        );
+        // Tell jest we're done
+        done();
```
tilleryd: ✨

src/components/Main/scenes/Fabric/FabricGrid.test.js
kaitmore: @tilleryd Agreed. I did start working on this but ended up abandoning. With no unique ID's, it becomes really hard to target these groupings. Ditto for the toolbar button comment below. Do you have any suggestions? Would it be best to delete these tests until we have a better system in place?

tilleryd: @kaitmore Our pair programming generated something bad for the code base so I'm definitely ok off setting this for a future date when we have a better solution across the board.

tilleryd:
@kaitmore LGTM

PR #914, opened 2017-11-08
Fix: Redirect on invalid instance

### LINKED ISSUE #879: Improve User Experience for an Instance of a Valid Microservice not existing
If a route has a microservice and version that matches but an invalid Instance ID, the app renders the Instance Dashboard (likely for the default runtime), starts polling without knowing a valid endpoint, and then errors out.

image

Desired Behavior:

* After the dashboard mounts, wait for the response from /services to learn if this microservice and version exist or not.
* If the microservice name and version and valid, check to see if the instance ID matches a value associated with the microservice name and version.
* If the instance ID matches, begin polling
* If the instance ID does not match, redirect back to the Fabric Dashboard and present a notification that the "489375340fsd87gf9-384qf4 is not a known instance of Service Name version 2.0";

kaitmore:
Resolves #879


src/components/Main/scenes/Fabric/FabricGrid.js
```diff
@@ -41,18 +41,10 @@ class FabricGrid extends Component {
...
-      // Reset location state
-      history.replace({
-        state: ""
-      });
-    } else if (state && !state.metered) {
-      notification(`${state.serviceName} does not have metrics to display`, {
+    // State added by fabric router
+    // Display message if one is found on state
+    if (state && state.message) {
+      notification(state.message, {
```
spmcbride1201: Can you please refactor to use the reportError utility function?

src/components/Main/scenes/Fabric/index.js
```diff
@@ -53,15 +53,31 @@ function FabricRouter({ services }) {
           const metered = service && service.metered;
           // Lookup the runtime of the microservice named serviceName
           const runtime =
-            services && serviceName && services[`${serviceName}|${version}`]
-              ? services[`${serviceName}|${version}`].runtime
-              : "";
-          // runtime informs the runtime-agnostic InstanceRouter which runtime router to render
-          // baseURL is prefixed to route paths and link to attributes when running with Fabric Server
+            services && serviceName && service ? service.runtime : "";
```
spmcbride1201: I'm thinking this could be a bit easier to understand if the variables are named to make it a bit more clear that they're booleans:
- userIsAuthorized
- serviceIsMetered
- instanceIsValid

spmcbride1201:
Looking back at the issue, I think I mistyped. The behavior for this scenario should be to redirect to the service dashboard (showing the list of instances).

kaitmore:
@spmcbride1201 @karinakeen 

- [x] Added version # to error message
- [x] An invalid instance now redirects to the service dashboard instead of Fabric Grid
- [x] Refactored notifications to use reportError utility function

PR #940, opened 2017-11-10
Better use of Button component

### LINKED ISSUE #795: Center Glyph in Icon component, regardless of Glyph scale ratio
Glyphs take a `ratio` property that scales the glyph within its Icon wrapper, but it does so based on the top-left corner. Glyphs need to be translated back to the center after scaling using either translate() or matrix().

shanberg:
- Transitioning StyledButton and derivatives to Button component
- Adding control over glyph ratio and color from Button component
- Adding missing Focus style properties to Button
- Additional minor tweaks to Button style 
- Closes #795


src/components/Button/Button.js
```diff
@@ -75,17 +74,14 @@ function Button({
...
-        <div width="100%" className="btn-glyph">
-          <svg width="100px" height="80px" viewBox="0 0 50 60">
-            <Glyph name={glyph} ratio={2} />
-          </svg>
-        </div>
+        <Icon>
+          <Glyph
+            glyphColor={glyphColor || "currentColor"}
+            name={glyph}
+            ratio={glyphRatio || "1"}
```
tilleryd: The default prop for ratio is 1 already so the OR expression isn't needed.

tilleryd: Unless glyphRatio is null or undefined...?  Which it shouldn't IMO.

src/components/Glyphs/index.js
```diff
@@ -123,7 +124,7 @@ export default class Glyph extends Component {
   };
 
   render() {
-    let { name, ratio } = this.props;
+    let { name, ratio = 1, glyphColor = "currentColor" } = this.props;
```
tilleryd: The idiomatic React pattern for setting default props is to use a React component's `defaultProps` property.

```
Glyph.defaultProps = {
  currentColor: "currentColor",
  ratio: 1
};
```

src/components/Button/Button.js
```diff
@@ -75,17 +74,14 @@ function Button({
...
-      )}
       {glyph && (
-        <div width="100%" className="btn-glyph">
-          <svg width="100px" height="80px" viewBox="0 0 50 60">
-            <Glyph name={glyph} ratio={2} />
-          </svg>
-        </div>
+        <Icon>
+          <Glyph
+            glyphColor={glyphColor || "currentColor"}
```
tilleryd: The Glyph component's default prop for glyphColor is `"currentColor"` already so the OR expression isn't needed.

src/components/Glyphs/index.js
```diff
@@ -135,11 +136,22 @@ export default class Glyph extends Component {
     // dynamically render glyph component by name
     const GlyphComponent = glyphs[name];
 
+    const viewBoxSize = 24;
```
tilleryd: You might want to move this to `src/utils/constants` and use it here as well as `src/components/Button/Button.scss` in place of `24px`.

src/style/mixins.scss
```diff
@@ -47,6 +57,25 @@ $button-font-weight: $button-font-weight-default) {
...
+  &:active {
+    transition: all 0 ease;
+  }
+
+  &:not[disabled] {
+    &:focus {
+      outline: none;
+      box-shadow: inset 0 0 0 rgba(255, 255, 255, 0.5),
+        0 0 0 1px transparentize($color-highlight, 0.25),
+        0 0 0 4px transparentize($color-highlight, 0.75);
```
tilleryd: @spmcbride1201 We're ultimately moving from SASS to styled-components right?  Should we stop SASS only functions from entering the code base now?  Or just refactor en mass when the time comes?

spmcbride1201: Our goal is to drop SCSS, but we haven't had the bandwidth to do it entirely. At this point, we've slimmed down and replaced most of the SCSS with styled-components, but there are a few components that still depend on it. I'm okay with pragmatic changes to the SCSS to fix UX issues for the release for existing components that depend on it, but anything beyond that seems to go against our goal to drop the CSS preprocessor.

src/style/mixins.scss
```diff
@@ -541,6 +570,12 @@ $button-font-weight: $button-font-weight-default) {
     }
   }
 
+  &:focus {
```
spmcbride1201: Since all of these mixins are only used in Button.scss, I'd like all of this logic to be move into that file and delete all the Sass imports of mixins.scss throughout the project.

src/components/Button/Button.js
```diff
@@ -75,17 +74,10 @@ function Button({
...
-          <Glyph name={icon} />
-        </Icon>
-      )}
       {glyph && (
-        <div width="100%" className="btn-glyph">
-          <svg width="100px" height="80px" viewBox="0 0 50 60">
-            <Glyph name={glyph} ratio={2} />
-          </svg>
-        </div>
+        <Icon>
```
spmcbride1201: This API is locking us into only using glyphs that are explicitly imported by glyphColor. The end user should always be able to manually pass in a Glyph and Icon as children so they can use custom glyphs.

src/style/mixins.scss
shanberg: Forms are also using this css, so we can't do that without breaking them, until we refactor the css into styled components.

PR #942, opened 2017-11-10
Editable Icon Border Widths

shanberg:
- Added `borderWidth` property to Icon border, passed into IconBorder component
- Removed excess elements from existing icon border elements to allow border widths to apply


src/components/Icon/components/IconBorder/IconBorder.js
```diff
@@ -18,30 +18,39 @@ export default class IconBorder extends Component {
...
+      <g
+        id={name}
+        className="iconBorder"
+        stroke={borderColor}
+        strokeOpacity={borderOpacity}
+        strokeWidth={borderWidth / ratio}
+        fill="transparent"
+      >
+        <IconBorderComponent name={name} />
+      </g>
```
kaitmore: This seems like a better pattern 👍🏼

PR #943, opened 2017-11-10
No More UIKit CSS

shanberg:
- Removes UIKit scss import from stylesheet
- Replaces the two styles we were rely on UIKit for


src/style/styleVariables.js
```diff
@@ -151,6 +151,8 @@ injectGlobal`
   #root {
     display: flex;
     flex-direction: column;
+    margin: 0;
+    padding: 0;
```
kaitmore: Nice, this fixes that weird flash of margin around the body as the page loads

PR #954, opened 2017-11-13
Fix: removed fixed height for table drawers

kaitmore:
Fixes #954


karinakeen:
Adjusting the width back to width solves the threads view, but cuts off x-axis in Routes graph view.

screen shot 2017-11-13 at 2 43 34 pm

karinakeen:
how graph looks on edge
screen shot 2017-11-13 at 2 45 56 pm

kaitmore:
@karinakeen Set chart-height in the scss to auto, and this seemed to fix it without causing problems elsewhere. Can you confirm that it doesn't break the chart anywhere else in the app?

karinakeen:
@kaitmore Works in Chrome. X-axis gets cut off in IE11 though. Approving this since I think IE11 issue was alive before the fix anyway...

screen shot 2017-11-13 at 3 57 41 pm
@kaitmore 


screen shot 2017-11-13 at 3 58 41 pm

kaitmore:
@karinakeen Ok, I can open up another issue for IE

PR #956, opened 2017-11-13
Fix: stop polling & clear metrics cache on invalid service or instance

kaitmore:


src/services/instance/metrics/index.js
```diff
@@ -108,6 +108,22 @@ function stopPollingInstanceMetricsEffect() {
 Effect("stopPollingInstanceMetrics", stopPollingInstanceMetricsEffect);
 
 /**
+ * Action that stops polling and clears metrics cache
+ */
+function stopPollingAnPurgeInstanceMetricsEffect() {
+  console.log("Polling stopped and metrics cache cleared");
```
7ruth: Just making sure that you did mean to leave this console here...

src/services/fabricMicroservices/fabricMicroservices.js
```diff
@@ -174,3 +174,12 @@ export function selectInstanceEffect({
     }
   }
 }
+
+/**
+ * Action that sets selected service, version, and instance to null
+ */
+export function deselectInstanceEffect() {
+  Actions.setSelectedService(null);
```
karinakeen: maybe rename function to clearInstanceEffect?

initial values for selectedService, selectedServiceVersion, and selectedServiceInstance are null.  you just pass nothing instead of passing null.

src/services/instance/metrics/index.js
```diff
@@ -108,6 +108,22 @@ function stopPollingInstanceMetricsEffect() {
...
+  Actions.stopPollingInstanceMetrics();
+  Actions.setMetricsPollingFailures(0);
+  Actions.clearMetrics();
+  Actions.deselectInstance();
+}
```
karinakeen: Instead of creating another Action called deselectInstance that calls three existing actions, maybe it's better to call on the three actions directly.  deselectInstance is only being used in stopPollingAnPurgeInstanceMetricsEffect and this effect already just calls on other effects.

karinakeen: Noticed typo
stopPollingAnPurgeInstanceMetricsEffect => stopPollingAn`d`PurgeInstanceMetricsEffect

src/services/fabricMicroservices/fabricMicroservices.js
kaitmore: @karinakeen I was passing null so as to be more explicit about what this was doing

src/services/instance/metrics/index.js
kaitmore: @karinakeen I thought that since there is a selectInstance effect, having a deselectInstance effect made sense semantically and could be useful at some point in the future, but if you don't think so I'll refactor

src/services/instance/metrics/index.js
kaitmore: Yep👍🏼

src/services/instance/metrics/index.js
karinakeen: going by the naming scheme, deselectInstance should mirror the existing selectedInstance and call on other actions such as stopPollingInstanceMetrics(), clearMetrics, startPollingInstanceMetrics() etc, which is already happening in stopPollingAnPurgeInstanceMetricsEffect(). I think the way it's set up now is stopPollingAnPurgeInstanceMetricsEffect() is a more comprehensive version of deselectInstanceEffect().

kaitmore: Ok, thanks, I'll refactor this

karinakeen:
looks great

PR #973, opened 2017-11-14
Test: GMServiceTableToolbar

### LINKED ISSUE #621: Service View: unit test: toolbar

kaitmore:
Resolves #621

## Proposed Changes

  - Slight refactor to GMServiceTableToolbar with the removal of unnecessary "serviceName" prop
  - Generated snapshot
  - Tested that all function props are being invoked properly
  
## How to Test

  - `npm run test`

## Notes 

 - It is difficult to test `react-select` (see [this issue](https://github.com/airbnb/enzyme/issues/400)). I needed to mount rather than shallow render `GMServiceTableToolbar` in order to have access to our `react-select` component. However, it seems that that `react-select` doesn't register a change event on the input when mounted. That's why I just called props.onChange() directly which in turn calls the mock function, `setSortByAttribute`


src/components/Main/scenes/Service/components/GMServiceTableToolbar.test.js
```diff
@@ -0,0 +1,46 @@
...
+import { mount, shallow } from "enzyme";
+import { MemoryRouter, Route } from "react-router";
+import GMServiceTableToolbar from "./GMServiceTableToolbar";
+import renderer from "react-test-renderer";
+import GMSelect from "../../../components/GMSelect";
+
+const mockData = {
+  setFilterString: jest.fn(string => console.log(`setFilterString was called`)),
+  setSortByAttribute: jest.fn(string =>
+    console.log(`setSortByAttribute was called`)
```
tilleryd: What purpose do these console.log statements serve? I wonder if they're necessary since we're only testing for the number of calls to the functions.

src/components/Main/scenes/Service/components/GMServiceTableToolbar.js
```diff
@@ -21,7 +20,6 @@ GMServiceTableToolbar.defaultProps = {
 };
 
 function GMServiceTableToolbar({
-  serviceName,
```
tilleryd: Nice catch! 🐟

src/components/Main/scenes/Service/components/GMServiceTableToolbar.test.js
kaitmore: I was just using them while developing the tests, I'll clean this up.

kaitmore:
@tilleryd I cleaned up imports/console logs

PR #986, opened 2017-11-15
feat: if no services are available under a specific heading, that sec…

### LINKED ISSUE #723: Fabric Page: Empty groupings should not be displayed
On the Fabric page in a Cards view, with Group: Status and Sort: Name, if the Down and Warning groups are empty, I think we should hide them from the view.

There are probably other things we should do in this case, but I can't think of them at the moment.

7ruth:
…tion is no longer displayed on fabric landing page

Resolves #723 

## Proposed Changes

  -Put an if test around the container that renders the section with microservices
  
## How to Test

  -There might be a better way (such as adjusting what sds is sending), but I hardcoded "" to dataGroupedByHeader['warning'] (for example) in src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/index.js around line 51 to simulate no microservices being passed under a specific heading.


src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/index.js
```diff
@@ -39,7 +39,6 @@ function SectionCardsView({
     const statusSort = pathname.slice(1);
```
kaitmore: Here I'm grabbing the status from the url bar

src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/index.js
```diff
@@ -39,7 +39,6 @@ function SectionCardsView({
...
 
     let headers;
-
     // If we pulled a status from the path, use that as the header
     // Else if we are on the main fabric grid, use microServiceStatuses as the headers
     if (statusSort && groupByAttribute === "Status") {
```
kaitmore: So I think you could probably get right of this particular check and simply check whether `groupByAttribute === "Status"`. We need to do that because the headers need to be sorted in a particular order for "Status" grouping.

src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/index.js
```diff
@@ -50,9 +49,24 @@ function SectionCardsView({
...
+    const verifiedHeaders = [];
+    for (let header of headers) {
+      // several checks to make sure there are microservices within the array
+      if (
+        !_.isEmpty(dataGroupedByHeader[header]) &&
+        typeof dataGroupedByHeader[header] !== "undefined" &&
+        dataGroupedByHeader[header] !== null &&
+        dataGroupedByHeader[header].length > 0 &&
+        dataGroupedByHeader[header][0] !== "" &&
+        dataGroupedByHeader[header][0] !== undefined
```
kaitmore: This seems like overkill to me, can we pare it down?

kaitmore: Also, perhaps you could use the `.filter()` array method here

src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/index.js
kaitmore: To clarify, I mean just keep the following checks:
```
if (groupByAttribute === "Status") {
       headers = microserviceStatuses.map(item => item.toLowerCase());
     } else {
        headers = headerTitles; 
}
```

src/components/Main/scenes/Fabric/components/FabricMainView/scene/Card/index.js
7ruth: In what way can I use .filter() here? (since it would still have to check for similar conditions) I will cut down 2 conditions which I deem redundant after closer examination.

kaitmore: @7ruth Oh I was thinking you could do something like:

```
headers.filter(header => {
return !_.isEmpty(dataGroupedByHeader[header]) // etc.
}
```

PR #988, opened 2017-11-15
fix: fixed slider css to be in line with mock ups

### LINKED ISSUE #966: Settings View slider is slightly off
## Expected Behavior
The bottom part of the slider should line up with the top part.

## Actual Behavior
No longer lines up

  1. Navigate to settings view and look at the slider

## Specifications

  - Version: 0.9.1
  - Platform: Fabric

7ruth:
Resolves #966 

## Proposed Changes

  -3rd party slider's css classes need to be over written to add styling we want

## How to Test

  -Navigate to Settings view and check out the sliders.


src/style/components/reactInputRange.css
```diff
@@ -1,8 +1,7 @@
-.input-range__track--active,
 .input-range__slider {
```
kaitmore: When I did a search for `input-range__slider`, it seems like there are actually already rules for it in Settings.scss. I believe you could just add these rules there, which will avoid having to use `!important` and allow you to delete this file altogether.

PR #989, opened 2017-11-15
637 settings view tests

7ruth:
Resolves #

## Proposed Changes

  -Adding tests for the Setting view and underlying PollingSettings component

## How to Test

  -Navigate to src/components/Main/components/Settings/index.test.js, remove 'x' before describe.
  -Navigate to src/components/Main/components/Settings/index.js and comment out ConfirmationModal
  -Run npm test

If the confirmation modal is not commented out, you will see this error "Enzyme Internal Error: unknown node with tag 4" which is due to react-modal using ReactDom.createPortal which is not yet supported in the upgraded version of Enzyme. Here is the PR that fixes the issue but not yet released: https://github.com/airbnb/enzyme/pull/1345

Description of the issue: https://github.com/airbnb/enzyme/issues/1150
 
Discussed this approach with David and he concurs that we can 'uncomment' that test at some point in the future when Enzyme is released and encompasses the feature that we need.


src/components/Main/components/Settings/components/PollingSettings.test.js
```diff
@@ -0,0 +1,95 @@
...
+describe("PollingSettings component", () => {
+  test("Matches the snapshot", () => {
+    const PollingSettingsComponent = renderer
+      .create(<PollingSettings {...mockProps} />)
+      .toJSON();
+    expect(PollingSettingsComponent).toMatchSnapshot();
+  });
+
+  test("Resume / Pause button component is rendered", () => {
+    const wrapper = mount(<PollingSettings {...mockProps} />);
```
kaitmore: Perhaps you could also put this wrapper in the beforeEach so you don't have to define it every test

src/components/Main/components/Settings/components/PollingSettings.test.js
```diff
@@ -0,0 +1,95 @@
...
+    expect(wrapper.find("button").length).toBe(1);
+  });
+
+  test("If isPolling prop is set to 'true', button displays 'Pause Updates'", () => {
+    mockProps.isPolling = true;
```
kaitmore: ...and then instead of this you can do `wrapper.setProps({isPolling: true})`

src/components/Main/components/Settings/components/PollingSettings.test.js
```diff
@@ -0,0 +1,95 @@
...
+        .find("button")
+        .find("g")
+        .at(1)
+        .hasClass("glyph")
+    ).toBe(true);
+  });
+
+  test("LayoutSection component is rendered by making sure header tag contains 'Instance Polling' text", () => {
+    const wrapper = mount(<PollingSettings {...mockProps} />);
+    expect(wrapper.find("header").text()).toBe("Instance Polling");
```
kaitmore: You could also check for the component display name: 
`expect(wrapper.find("LayoutSection")).toHaveLength(1);`

src/components/Main/components/Settings/index.test.js
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  test("Has Cache Size readout", () => {
+    SettingGridWrap = mount(<SettingsGrid store={mockStore(mockState)} />);
+    expect(SettingGridWrap.find("Readout").props().readoutItems[0].title).toBe(
+      "Cache Size"
+    );
+  });
+
+  test("Clear Cache button exists", () => {
```
kaitmore: I think this test could be simplified by using an object property selector like: `expect(SettingGridWrap.find({ title: "Clear Cache" })).toHaveLength(1)`

src/components/Main/components/Settings/index.test.js
```diff
@@ -0,0 +1,75 @@
...
+xdescribe("SettingsGrid component", () => {
+  test("Matches snapshot", () => {
+    const SettingGrid = renderer
+      .create(<SettingsGrid store={mockStore(mockState)} />)
+      .toJSON();
+    expect(SettingGrid).toMatchSnapshot();
+  });
+
+  test("Has an errorBoundary", () => {
+    SettingGridWrap = mount(<SettingsGrid store={mockStore(mockState)} />);
```
kaitmore: Same, you should probably mount in the beforeEach

src/components/Main/components/Settings/index.test.js
```diff
@@ -0,0 +1,75 @@
...
+import thunk from "redux-thunk";
+import { Provider } from "react-redux";
+
+import SettingsGrid from "./index";
+
+//import all necessary Action effect so the Actions object is set within this test harness
+import "services/fabricMicroservices/index.js";
+import "services/instance/metrics/index.js";
+
+const middlewares = [thunk],
```
kaitmore: I think you can get rid of the thunk middleware

src/components/Main/components/Settings/index.test.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  test("Has 2 Fabric Polling Settings if fabricServer prop is available", () => {
+    mockState.settings.fabricServer = "http://localhost:1337";
+    SettingGridWrap = mount(<SettingsGrid store={mockStore(mockState)} />);
+    expect(SettingGridWrap.find("PollingSettings").length).toBe(2);
+  });
+
+  test("Has Cache Size readout", () => {
+    expect(SettingGridWrap.find({ title: "Clear Cache" })).toHaveLength(1);
```
kaitmore: I was thinking that this expect would replace the "Clear Cache button exists" test below. find({title: "Clear Cache"}) is actually targeting the button

screen shot 2017-11-16 at 1 40 20 pm

spmcbride1201:
I think the portal support is in enzyme-adapter-react-16 v1.1.0. I'm going to upgrade and test to see if that's indeed the case.

src/components/Main/components/Settings/components/PollingSettings.test.js
```diff
@@ -0,0 +1,86 @@
+import React from "react";
+import { shallow, mount } from "enzyme";
```
spmcbride1201: Looks like shallow isn't used. Please remove zombie imports.

src/components/Main/components/Settings/components/index.js
```diff
@@ -0,0 +1,39 @@
+import { Effect } from "jumpstate";
```
spmcbride1201: I'm really confused by this file. It looks like it's unused and was accidentally copied from `/src/services/fabricMicroservices/index.js`. Was this placed here intentionally for some reason?

src/components/Main/components/Settings/index.test.js
spmcbride1201: Looks like he did this.

src/components/Main/components/Settings/index.test.js
```diff
@@ -0,0 +1,62 @@
...
+//import all necessary Action effect so the Actions object is set within this test harness
+import "services/fabricMicroservices/index.js";
+import "services/instance/metrics/index.js";
+
+const mockStore = configureStore();
+
+let SettingGridWrap,
+  mockState = state.default;
+
+xdescribe("SettingsGrid component", () => {
```
spmcbride1201: I validated that this can stay uncommented if you update Enzyme and the adapter as follows:
```
"enzyme": "^3.2.0",
"enzyme-adapter-react-16": "^1.1.0",
```

Please update these dependencies, uncomment all tests, and verify that the app is able to run tests and start as expected.

src/components/Main/components/Settings/components/index.js
7ruth: I didn't clean it up, this was one of my attempts to get Actions imported. Will be deleted

PR #1026, opened 2017-11-16
Fix: Prevent Default Behavior when Hitting Enter in a Search Box

### LINKED ISSUE #992: Fabric View: initial search causes a full page refresh
## Expected Behavior
Typing a keyword search and hitting enter updates the URL and filters the services. 

## Actual Behavior
Typing a keyword search and hitting enter causes a full page refresh.

  1. Navigate to https://edge.deciphernow.com
  1. Type keyword search and hit enter
  1. Observe the full page refresh and the URL changes to https://edge.deciphernow.com/?#/
  1. Type keyword search and hit enter
  1. Observe that the page does not refresh

Conclusion: the issue is related to the URL changes. 

## Specifications

  - Version: 0.9.1
  - Subsystem: Safari 10.1

kaitmore:
Resolves #992 

## Proposed Changes

  - Added `e.preventDefault` to all form elements within toolbars across the app

## How to Test

  - Navigate to toolbars with search boxes: Fabric View, Service View, JVM -> Routes, Go -> Routes, Functions 
  - On each view, focus the search box and hit enter


kaitmore:
@tilleryd Noticed this bug applied to Chrome as well. Let me know if this solves it on Safari 10

tilleryd:
@kaitmore Yup this fixes the issue in Safari an Chrome.  What this solution implies is that we need to have a `Form` component.  That way it will help enforce future uses of form elements (as best as possible).  Thus when a developer uses the Form component they get the `preventDefault` behavior automatically.

tilleryd:
@kaitmore You'll also need to update Jest snapshots for some of the tests that are failing.

PR #1033, opened 2017-11-16
Fix: Toggle asc/desc Order in Service View

### LINKED ISSUE #1023: Fabric Main View and JVM/GO List of Instances View - Selecting same sort option should toggle ascending/descending order
## Expected Behavior

Selecting same sort option[Name, Status] should toggle ascending/descending sort order 

## Actual Behavior

Doesn't change ascending/descending order

For Main Fabric View
  1.  On main fabric view, choose either Card or List view, and click on the same Sort option twice

For JVM/GO service list of instances view
  1.  Navigate to either JVM/GO service, and click on the same Sort option twice 

  
screen shot 2017-11-16 at 11 28 26 am

screen shot 2017-11-16 at 11 51 15 am

## Specifications

  - Version: macOS High Sierra V 10.13
  - Platform: Witnessed in Chrome 62, Safari 11.0, expecting the same behavior in all other browsers
  - Subsystem:

kaitmore:
Relates to #1023 

## Proposed Changes

  - Enhanced `setSortByAttribute` instance method to handle toggling asc/desc order

## How to Test

  - Navigate to a JVM or GO service view
  - Sort instances by name, and then make the same selection. This should toggle the instances from asc to desc.


tilleryd:
@kaitmore I tested this PR and indeed the bug is fixed but it only addressed the Service instance view.  Issue #1023 states that it's happening on the Fabric view as well.  You can address the fix in this PR or follow up with another.  In either case I think `setSortByAttribute` can be a utility function used by both toolbars.

src/components/Main/scenes/Service/service.test.js
```diff
@@ -63,7 +64,36 @@ describe("Service View", () => {
...
 
-  test("sorts by name", () => {
+  test("sorts by start time in desc order", () => {
+    let instance = wrapper.find("GMServiceView").instance();
+
+    // Setting the sortByAttribute twice should reverse the order
+    instance.setSortByAttribute("start_time");
+    instance.setSortByAttribute("start_time");
+
+    const positionOf_1ridkelvyaw = wrapper.html().indexOf("1ridkelvyaw");
```
tilleryd: I know you've been using these IDs for previous tests but they're hard to read.  Can you just use simple names like "serviceID1", "serviceID2", and "serviceID3"?  Since we're not actually testing the IDs (and don't need to) and we're not trying to emulate reality 1:1, I think legibility in the tests is more beneficial.

kaitmore:
@tilleryd All service names have been updated in tests to be more readable

PR #1039, opened 2017-11-16
Rebuild node-sass

tilleryd:
CircleCI is caching dependencies with 7.x and we just upgraded to 8.x.  There doesn't seem to be a good way of invalidating the cache without human intervention through the UI.  By rebuilding `node-sass` we can mitigate this issue.  I'm not comfortable with this as a long term solution but it does resolve the immediate problem.


tilleryd:
@spmcbride1201 Your fixes appear to be working.  LGTM

PR #1093, opened 2017-11-27
fix: floatRound for sparklineNetChange

### LINKED ISSUE #1008: Instance view: route sparklines do not always show large spikes that are shown in the large chart
screen shot 2017-11-15 at 5 09 34 pm

Looking at the screenshot's `/homehtml` chart, it seems that 0 activity followed by a spike then followed by 0 activity again may be the culprit.  The `/metrics_service_versions_inst_` chart ad a lot of activity with 1 spike which is displayed in it's associated sparkline.  

## Specifications

  - Version: 0.9.1
  - Subsystem: Safari 10.1

karinakeen:
Resolves #1008 

## Proposed Changes

  - `getSparkLineOfNetChange` computes net change as integers for sparkline displays. `_netChangeMapper` computes net change as floating points for charts. this PR implements floatRound in place of Math.round in `getSparkLineOfNetChange` to keep the values consistent. 
  
## How to Test

  -
  -
  -


src/utils/sparklines.js
```diff
@@ -32,11 +32,11 @@ export function getSparkLineOfNetChange(metrics, key) {
...
-      return idx === 0
-        ? 0
-        : Math.round(
-            (Number(value) - Number(last)) / ((currentTime - lastTime) / 1000)
-          );
+      let netChange = floatRound(
+        (Number(value) - Number(last)) / ((currentTime - lastTime) / 1000),
+        3
+      );
+      return idx === 0 || netChange < 0 ? 0 : netChange;
```
kaitmore: So if idx === 0, then won't it just return `true`?

karinakeen: if netChange is negative, netChange should be zero instead of negative value

PR #1096, opened 2017-11-27
fix: transform scale on hover of service cards no longer causes a tex…

### LINKED ISSUE #990: Service cards are blurry on hover in Safari
## Expected Behavior
Service cards enlarge slightly on mouse hover, appearing otherwise normal.

## Actual Behavior
Service cards enlarge slightly on mouse hover and appear blurred.

1. Open the app at edge.deciphernow.com.
2. Hover over the card for a stable service.
3. Note that the contents of the card are blurry.

## Specifications
- Version: 0.9.1
- Platform: Safari 11.0.1 on macOS High Sierra 10.13.1

7ruth:
…t blur in safari

Resolves #990

## Proposed Changes

  -Adding backface-visibility: hidden; and -webkit-font-smoothing: antialiased; 
  -Combination of those produced the best results in safari (and other browsers) in terms of text being least distorted on hover

  
## How to Test

  -In safari, navigate to card view and hover over a card


karinakeen:
I tested this on Safari, compared the change with Edge side by side, at different zoom levels.  I am not able to determine if one is better or worse.  The texts narrow slightly compared to edge but can't say  if it's less or more blurry.  @7ruth Did adding font-smoothing to other browsers make it better or worse?

7ruth:
@karinakeen so I also tried messing with font-smooth, auto didn't work well... made worse in my opinion, increased the blur. The narrowing I see now, gives it more definition.

7ruth:
@jsweetland , could you please review this change, I was able to get it slightly less blurry by enforcing antialiasing. One other thing that worked, but the gain in fidelity was hard to see, at least it didn't make it more blurry, was removing h1 on line 70 in GMServiceCard/components/Card.js, but that gets rid of transform effect on the text.

tilleryd:
@karinakeen Please review this PR.  We don't need to pull in @jsweetland.

karinakeen:
@7ruth instead of applying scale to text, try increasing font size on hover?

7ruth:
@karinakeen , on hover, the scale of the h1 actually goes down... the formula is 1/${CARD_INTERACTION_SCALE}. So that way when you hover, size of the card grows but text size supposedly remains the same (except for fuzziness). 

I tried adjusting font-size smaller on hover, it resizes the text within the card (looks weird) and doesn't produce any sharper text (also makes it a bit fuzzy in the process). 

Trying out other solutions people proposed on the forums and I found something that works slightly better then what I initially proposed. Check out latest commit to this PR. I added translate3d(0,0,0); to the transform property and still makes it a slightly fuzzy at large zooms but significantly less than before (letters no longer move in their positions at high zoom).

karinakeen:
@7ruth I am not being picky but I see zero difference between applying translate3d(0,0,0) and doing it without.  I tested on higher and lower zoom and I still don't think this is any improvement for the blurred text on safari browser. @kaitmore maybe second set of eyes?

7ruth:
screen shot 2017-11-28 at 6 10 35 pm

screen shot 2017-11-28 at 6 11 09 pm

karinakeen:
changing my review per confirmation by @kaitmore

Issue #610, opened 2017-10-18
JVM Threads Table - Implement new toolbar format shared across app


spmcbride1201:
![image](https://user-images.githubusercontent.com/8700322/32116009-e94127b6-bb16-11e7-8643-20cda3f67349.png)

PR #1099, opened 2017-11-28
Feat: Reusable Table Toolbar Component

### LINKED ISSUE #992: Fabric View: initial search causes a full page refresh

### LINKED ISSUE #1021: Instance view: Go: Routes: sorting by Error % only works once
## Expected Behavior
Sorting by Error % should work consistently not just once.

## Actual Behavior
Sorting by Error % works once but then subsequent attempts do nothing.

  1. Navigate to a Go service instance
  1. Click on the Routes tab
  1. Sort by Error %
  1. Sort by Error % again
  1. Observe that sorting doesn't take place

## Specifications

  - Version: 0.9.1
  - Subsystem: All browsers

### LINKED ISSUE #1042: Reusable table toolbar component
- [ ] Create Reusable table toolbar #1069 
- [ ] Implement Unit Tests #1067 
- [ ] Implement Storybook Stories #1068

### LINKED ISSUE #1067: Unit tests: reusable table toolbar component
Relates to #1042

### LINKED ISSUE #1068: Storybook Stories: reusable table toolbar component
Relates #1042

### LINKED ISSUE #1069: Create reusable table toolbar component
Relates #1042 

- [ ] Create Reusable table toolbar
- [ ] Implement across all table instances
- [ ] Delete `forms.scss`

kaitmore:
Resolves #1067 
Resolves #1068 
Resolves #1069 
Resolves #1042 
Resolves #1021 
Resolves #992 

## Proposed Changes

  - There is now one TableToolbar that renders the appropriate subcomponents (search, display toggles, dropdowns) based on the props it receives. It is also possible to bypass these standard subcomponents altogether by using the toolbarChildren props to pass any custom element or node to the corresponding column.
- Deleted old toolbars and tests
- Added new unit test
- Added story (_Note: I still can't get styles to load for GMSelect.js and Button.js in storybook_)

## How to Test
- run `npm start` and verify the toolbars render and function as expected in the following areas:
    - Fabric Main View
    - Fabric Status Views (Stable, Warning, Down)
    - Service View
    - Go -> Functions & Routes
    - JVM -> Routes
 _Note: ThreadsTable was not touched, as it needs to be further refactored (see #610)_
- Type something into any toolbar search input and hit enter. Verify that the page does not hard refresh.
- run `npm test` and verify TableToolbar tests pass
- run `npm run storybook` and verify TableToolbar story renders


src/components/Main/scenes/FabricView/components/FabricGrid/FabricGrid.js
```diff
@@ -211,5 +226,34 @@ class FabricGrid extends Component {
...
+  {
+    value: "Owner",
+    label: "Owner"
+  },
+  {
+    value: "Capability",
+    label: "Capability"
+  },
+  {
+    value: "Status",
```
karinakeen: Maybe we shouldn't display groupBy Status when we are already on Fabric Status Views (Down/Warning/Stable View)
screen shot 2017-11-28 at 9 42 01 am

src/components/Main/components/TableToolbar/TableToolbar.js
```diff
@@ -0,0 +1,131 @@
...
+    groupByAttribute: PropTypes.string.isRequired, // value for controlled GMSelect Group By dropdown
+    groupByOptions: PropTypes.array.isRequired, // array of option objects for the GMSelect Group By dropdown
+    setGroupByAttribute: PropTypes.func.isRequired // onChange handler for GMSelect Group By dropdown
+  }),
+  searchInputProps: PropTypes.shape({
+    filterString: PropTypes.string.isRequired, // value for controlled SearchInput input
+    setFilterString: PropTypes.func.isRequired, // onChange handler for search input
+    searchPlaceholder: PropTypes.string.isRequired // placeholder for SearchInput
+  }),
+  sortByProps: PropTypes.shape({
```
karinakeen: sortByProps is set to false initially and throws warning
screen shot 2017-11-28 at 10 08 32 am

src/components/Main/components/TableToolbar/TableToolbar.js
```diff
@@ -0,0 +1,131 @@
...
+              onChange={evt =>
+                searchInputProps.setFilterString(evt.target.value)
+              }
+              placeholder={searchInputProps.searchPlaceholder}
+              aria-label={searchInputProps.searchPlaceholder}
+              value={searchInputProps.filterString}
+            />
+          </Form>
+        )}
+        {toolbarLeftChildren}
```
karinakeen: what are the optional props being passed as children to toolbarleft, -center,- right?

src/components/Main/components/TableToolbar/TableToolbar.js
kaitmore: Good catch, I updated the logic for passing sortByProps in status view.

src/components/Main/components/TableToolbar/TableToolbar.js
```diff
@@ -0,0 +1,131 @@
...
+      <ToolbarRight>
+        {groupByProps && (
+          <GMSelect
+            name="form-field-group-by"
+            options={groupByProps.groupByOptions}
+            value={groupByProps.groupByAttribute}
+            onChange={val => groupByProps.setGroupByAttribute(val.value)}
+            clearable={false}
+            searchable={false}
+            valueRenderer={val => (
```
karinakeen: bring valueRenderer out?

src/components/Main/components/TableToolbar/TableToolbar.js
kaitmore: Nothing right now, but my thinking was that if we ever encountered an edge case where we need a toolbar but wanted to replace the search input with something else, this could be a convenient escape hatch - e.g., we can just ignore passing in searchInputProps so that isn't rendered, and pass in a button instead. 

However, I just realized that as it stands, this wouldn't work for toolbarCenterChildren since the `displayTypeProps` check should be outside of <ToolbarCenter>.

karinakeen:
Confirms search and sort functions work and storybook runs.  Fails settingsgrid.test.js.  Please see other comments

src/components/Main/scenes/FabricView/components/FabricGrid/FabricGrid.js
kaitmore: Sean and I talked about this a while back, however since this logic is coupled with the FabricGrid, it becomes deceptively tricky to implement and I think it's out of the scope of this PR. We can make a followup issue to discuss it further.

src/components/Main/components/TableToolbar/TableToolbar.js
kaitmore: @spmcbride1201 Yup, I had already fixed it. Now the problem is that I had to wrap the buttons in that div, hence the weird CSS > div selector you noticed in the ToolbarCenter component, which is also what broke ThreadsTable. 😬

kaitmore:
@spmcbride1201 Addressed the following issues:

- _Fabric Status Views, when List and Status is selected, the list does not render properly_
**Fixed magically when I merged latest master, and now I can't recreate the problem. Let me know if it persists for you.**
- _On JVM Threads, The toolbar isn’t centered any longer._ 
**Fixed by updating the CSS for ToolbarCenter and getting rid of the weird div selector**
- **Created a** `GMSelectValueRenderer` **stateless functional component**
- **Updated all sort and group by options to be class properties. I needed to be able to access them in the instance itself, and to do that I would need to use** `FabricGrid.sortByOptions` **instead of** `this.sortByOptions`, **which felt weird to me, but maybe it's fine? See [this issue](https://github.com/gaearon/babel-plugin-react-transform/issues/98). Anyways, that's why I dropped the** `static`
- **Updated snapshots**


As for storybook, yeah, the dropdowns look weird because it's not importing the css file in `GMSelect.js`.

PR #1100, opened 2017-11-28
fix: if a status category doesnt receive microservices its section wi…

### LINKED ISSUE #723: Fabric Page: Empty groupings should not be displayed

7ruth:
…ll no longer be visible in the list view

Resolves #723 

## Proposed Changes

  -Adding a check to see if microservices are present under a heading key, if not, that section should not be rendered in the list view
  
## How to Test

  - /FabricMainView/scene/ServicesListView/ServicesListView.js
  - around line 48, add something similar to dataGroupedByHeader['down']='';
  - Navigate to list view after, no empty section should be visible.


src/components/Main/scenes/FabricView/components/FabricGrid/components/FabricMainView/scene/ServicesListView/ServicesListView.js
```diff
@@ -31,23 +31,29 @@ function ListView({
...
-    } else if (!statusSort && groupByAttribute === "Status") {
+    // If we are sorting by status, use microServiceStatuses as the headers
+    // as they need to be in the order "Down, Warning, Stable"
+    if (groupByAttribute === "Status") {
       headers = microserviceStatuses.map(item => item.toLowerCase());
     } else {
       headers = headerTitles;
     }
 
+    let verifiedHeaders = [];
```
kaitmore: You can delete this line and initiate the variable on line 46 
`let verifiedHeaders = headers.filter...`

PR #1122, opened 2017-11-30
1110 fix flex layout

### LINKED ISSUE #1110: Vitals overlap the GMLineChart when vitals reflow as screen width < 800px
## Expected Behavior

The vitals and the GMLineChart are on the same z-index, and when the vitals reflow, the GMLineChart shifts downwards with the same vertical padding at wider screen widths

## Actual Behavior

The vitals reflow and seem to be at a higher z-index, which causes them to reflow into the GMLineChart.

  1. Start app and open in IE11
  2. Open an instance dashboard
  3. Shrink the width of the window to < 800px to cause the vitals to reflow into multiple rows
  
image

## Specifications

  - Version: Master as of 11/28/2017
  - Platform: Windows 7/8/10
  - Subsystem: IE11

karinakeen:
Resolves #1110 

## Proposed Changes

  - GO/JVM Summary View has `AppContent` as parent container.  `AppContent` has flex setting of `1 1 100%`, changing this to `1 1 auto` would have solved the issue, but discussed with Stuart and instead added new flex setting `flex: 0 0 auto` to `LayoutSectionWrap` .
  - Note:`AppContent` retains `1 1 100%` to provide layout for the footer.
  - Note:`LayoutSectionWrap` is a parent container for both summary sections, `Vitals` and `Statistics`.
  
## How to Test

  - Navigate to JVM/ GO instance view and resize screen to <800px
  - tested on IE11, Chrome 62, Safari 11, Firefox 55, 56, 57, Opera 49


karinakeen:
On IE11 after the fix 
screen shot 2017-11-30 at 12 35 28 pm

PR #1130, opened 2017-12-01
Refactor: Notification SCSS

### LINKED ISSUE #1046: Refactor Notification from scss into styled-components

kaitmore:
Resolves #1046 

## Proposed Changes

  - There isn't really any way to actually refactor this SCSS into styled-components, as we need to target react-notification-system's classnames in order to style them. Instead, I broke out the SCSS into chunks and put them in their own style folder as sort of faux-styled-components. Until injectGlobal is factored out of the styleVariables file, I had to add another injectGlobal in `Notification.js`.
  
## How to Test

  - Don't change the index.html meta tag and make sure the notifications are rendered and styled properly.


src/components/Notification/Notification.js
```diff
@@ -1,7 +1,20 @@
...
+  NotificationTitle
+} from "./style";
+
+// Inject our "faux" styled-components that contain
+// 3rd party classnames into the global stylesheet
+injectGlobal`
+${NotificationWrapper};
+${NotificationBody};
+${NotificationTitle}
+`;
```
7ruth: String refs below should be eventually refactored to a callback. (Kait mentioned that she tried to refactor it, however it was causing other issues that were outside scope, good place to revisit later).

PR #1134, opened 2017-12-01
569 reusable gm table molecule

### LINKED ISSUE #569: Create a reusable GMTable molecule
Currently the .routes-table- prefixed styles are used in:

Fabric Table
Go Functions Table
Go Routes Table
JVM Routes Table
There is also a JVM Thread Table that might use distinct styles, but it very similar.

Identify the possibility of creating a common GMTable molecule that is sufficiently extensible to cover each of these use cases.

karinakeen:
Resolves #569 

## Proposed Changes
  - implements common table component for Go functions and routes table, JVM routes table, and GM Service table
  - deletes obsolete tables and child line item components
  - refactor old tests 
  - I did not include threadsTable as it will be refactored heavily, blocked by #614 
  - renamed Table styled-component to TableDisplay 
  - rename 'totalRequests' field to 'Requests' for JVM routes
  
## How to Test

  - navigate to JVM > Routes
  - navigate to Go > Routes
  - navigate to Go > Functions
  - navigate to both GO/ JVM Instance Service view


src/components/Main/components/Table/Table.js
```diff
@@ -0,0 +1,148 @@
...
+import React from "react";
+import _ from "lodash";
+
+import TableLineItem from "./components/TableLineItem";
+import GMServiceTableLineItem from "./components/GMServiceTableLineItem";
+
+import TableDisplay from "components/Main/components/TableDisplay";
+import TableHeader from "components/Main/components/TableHeader";
+import TableColHeader from "components/Main/components/TableColHeader";
+import TableBody from "components/Main/components/TableBody";
```
kaitmore: You could move these to the Table/components directory

src/components/Main/components/Table/Table.js
```diff
@@ -0,0 +1,148 @@
...
+  // apply greater flex-basis for first column (15% -> 20%)
+  // requests/sec is a sparkline header and set to align left
+  // set text-align to right for additional text columns
+  return headerNames.map((header, headerIdx) => {
+    styleProps = {};
+    if (headerIdx === 0) styleProps = { flex: "1 1 20%" };
+    if (headerIdx !== 0 && header !== "Requests/sec")
+      styleProps = { textAlign: "right" };
+    return (
+      <TableColHeader style={styleProps} key={`${header}{headerIdx}`}>
```
kaitmore: headerIdx is a string here, maybe do something like ${header|headerIdx}

src/components/Main/components/Table/Table.js
```diff
@@ -0,0 +1,148 @@
...
+
+import TableDisplay from "components/Main/components/TableDisplay";
+import TableHeader from "components/Main/components/TableHeader";
+import TableColHeader from "components/Main/components/TableColHeader";
+import TableBody from "components/Main/components/TableBody";
+
+import { relativeReqPercent, convertMS } from "utils";
+
+Table.propTypes = {
+  categories: PropTypes.array,
```
kaitmore: I'm confused about the term "categories". Maybe something like "items" would make more sense.

src/components/Main/components/Table/components/TableLineItem.js
```diff
@@ -7,34 +7,34 @@ import {
...
 import GMLineChart from "components/Main/components/GMLineChart";
-import SparklineCol from "components/Main/components/SparklineCol";
+import TableRow from "components/Main/components/TableRow";
 import TableCol from "components/Main/components/TableCol";
 import TableColVizBar from "components/Main/components/TableColVizBar";
+import SparklineCol from "components/Main/components/SparklineCol";
 import TableDrawerCollapse from "components/Main/components/TableDrawerCollapse";
-import TableRow from "components/Main/components/TableRow";
 import VizBar from "components/Main/components/VizBar";
 import VizFill from "components/Main/components/VizFill";
```
kaitmore: Some of these could also probably be moved to Table's component directory, unless they're being used elsewhere

src/components/Main/components/Table/components/TableLineItem.js
```diff
@@ -7,34 +7,34 @@ import {
...
+ * A row of data in Functions and Routes table
  * @export
- * @class RoutesTableLineItem
+ * @class TableLineItem
  * @extends {Component}
  */
-export default class RoutesTableLineItem extends Component {
+export default class TableLineItem extends Component {
   static propTypes = {
+    category: PropTypes.string.isRequired,
```
kaitmore: Same comment as before with the category prop name

src/components/Main/components/Table/Table.js
```diff
@@ -0,0 +1,148 @@
...
+ * takes categories and returns name of category as "Function", "Route" or ""
+ * @param {object} categories
+ * @returns {string} - "Function", "Route" or ""
+ */
+const getCategoryName = categories => {
+  if (_.isEmpty(categories)) return "";
+  if (_.has(categories[0], "func")) return "Function";
+  else if (_.has(categories[0], "route")) return "Route";
+  else return "";
+};
```
kaitmore: If you're checking this anyways, you probably don't need to have different "categories" and "instances" props, you could just have a single "items" array. Another idea: you could skip this helper func altogether and have Table accept a "type" prop which can either be "function", "route", or "instance", then decide which Line Item to render based on that.

src/components/Main/components/Table/Table.test.js
```diff
@@ -0,0 +1,150 @@
...
+import { mount } from "enzyme";
+import renderer from "react-test-renderer";
+
+import Table from "./index.js";
+import TableLineItem from "./components/TableLineItem";
+import GMServiceTableLineItem from "./components/GMServiceTableLineItem";
+import TableColHeader from "components/Main/components/TableColHeader";
+
+let wrapper;
+let routesProps = {
```
kaitmore: Tests look great! However I think you should test for when Table is passed instances and renders GMServiceTableLineItem

karinakeen:
Right now GM tables have pre-set styling based on the nature of each column element.  There are mainly three; one with name and vizbar, one showing sparkline, and the other plain text column.  If we were to make a fully dynamic table component that takes array of headers and items, we'd have to pass in styling notation.   May be out of scope for this specific issue.  Let's add it to discussion.

src/components/Main/components/Table/Table.js
karinakeen: I am passing in additional "Type" prop from parent containers now.  I rewrote the helper function to evaluate this "Type" prop.

src/components/Main/components/Table/components/TableLineItem.js
karinakeen: Some of theses are exported as extended styles so they can live on main until threadsTable doesn't need them.

src/components/Main/components/Table/Table.js
```diff
@@ -0,0 +1,148 @@
...
+  serviceVersion: PropTypes.string,
+  status: PropTypes.string
+};
+
+/** Renders JVM > Routes table, Go > Functions table, Go > Routes table, Instance Service table
+ * Includes a Header and maps 0..n rows of functions data to TableLineItems
+ * @export
+ * @param {any}
+ * @returns
+ */
```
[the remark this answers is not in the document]
karinakeen: moved

src/components/Main/components/Table/Table.test.js
karinakeen: instances test added

karinakeen:
@kaitmore implemented suggested changes.

PR #1176, opened 2017-12-05
fix: 1115 readout items are broken in storybook

### LINKED ISSUE #1115: Readout Items are broken in Storybook
## Expected Behavior
Readout Items should not overflow their containers

## Actual Behavior
Readout items are not rendering correctly in storybook

  1. `npm run storybook`
  2. Select any of the Readout stories
  
screen shot 2017-11-29 at 9 57 51 am

## Specifications

  - Version: 0.9.1
  - Platform: MacOS 
  - Subsystem: Chrome 62

karinakeen:
Resolves #1115 

## Proposed Changes
  - rename readout styled components to be more descriptive
  - move the custom styled props into styled components for readout items
  - fix props being passed to readout.stories.  currently we are not displaying sparklines as part of readout items.   this will be changed in the future. 
  
## How to Test

  - npm run storybook
  - check Readout section and confirm that readout components are rendering correctly


kaitmore:
It looks better in storybook, but I'm not sure if I agree about removing custom style props. Before we had it so that the one outlier case (SettingsGrid) could override a couple default styles. Now Readout has a super specific prop and no way to customize for other use cases. We would have to add another custom prop if we needed to reuse this and that can start to get crazy. The custom style props pattern that is currently in place was modeled off of material UI: http://www.material-ui.com/#/components/card

shanberg:
Can I propose a different solution?

The only reason we need different styling on the settings cache readout, as far as I can tell, is that it contains a button. It looks to me like all we'd have to do is support the layout of a button in a ReadoutItem for any readout, and then we wouldn't need either the dangerous style prop or the overly-specific cacheCard prop.

karinakeen:
My thought was if readoutItem won't have custom styles after we implement button module for cache card, we can just get rid of custom props `cachecard` from readout after button work is done.

kaitmore:
It was also because the cache readout needed to override the default flex property. We have a media query for widths < 1000px applying to all Readout containers, which didn't make sense for that use case.

karinakeen:
Metrics cache card doesn't look pixel perfect reusing readout component.  I can revert the containerStyle back if we are going down the path of reusing readout function to render all card-like items.

karinakeen:
We are going to disconnect the cache display from the readout component, as they are not really related.  This will be done in a separate issue.

PR #1196, opened 2017-12-08
1045 refactor button into styled components

### LINKED ISSUE #1045: Refactor Button from scss into styled-components
- [x] Refactor Button.scss into styled-components
- [x] Refactor NavButton.js and Button.js to use the styled-components
- [x] Delete zombie StyledButton 
- [x] Clean up unused mixins, scss functions, scss variables.

Note: `Inspector.scss` and `forms.scss` also uses button functions
```scss
  @include function-button-base();
  @include function-button-style();
  @include function-button-size();
```

7ruth:
Resolves #1045 

## Proposed Changes

  -Refactored button.scss and mixin.scss functionality into ButtonWrap component
  
## How to Test

  -Look at the buttons on the front end ( toggle Card/List buttons, buttons in the Settings view)


7ruth:
Just leaving a note here for later. In the future, we should decouple buttonOutlineStyles which are generated within functionButtonStyle. Also there is an opportunity to build few interfaces for generating buttonOutlineStyle type styles.

karinakeen:
@7ruth What are your thoughts on building an interface on generating styles?

7ruth:
Yeah, I think thats exactly what should be done next. 

There needs to be a conversation with Stuart regarding the knobs that he needs. I’d love to hear your thoughts on the design/approach, each individual button style should eventually live in storybook and you know ins and outs of that piece.

src/components/Button/components/ButtonWrap.js
```diff
@@ -0,0 +1,800 @@
+import styled from "styled-components";
```
spmcbride1201: This component really needs to be refactored into different files in order to better separate out utility functions. I'd rather get this merged in first and then deal with this though.

spmcbride1201:
Feel free to merge without addressing the Code Climate issues. We'll assume the technical debt for now.

PR #1197, opened 2017-12-11
WIP feat: 1178 reusable card component

### LINKED ISSUE #1178: Create a reusable card component that takes containerStyle props
Create a reusable card component that takes containerStyle props following material-ui pattern. 
 This component will be used for metrics cache card and similar cards in the future.

karinakeen:
Resolves #1178 

## Proposed Changes

  - Create a reusable card component that takes containerStyle props following material-ui pattern.
  - Card component is used for metrics cache card under settings page.
  - Remove cacheCard style from Readout
  - add styled-component test
  
## How to Test

  - Navigate to settings and confirm cache card is rendered correctly
  - run cache card story


src/components/Main/components/Card/Card.js
```diff
@@ -0,0 +1,73 @@
...
+import CardBody from "./components/CardBody";
+import CardChildren from "./components/CardChildren";
+
+export default function Card({
+  bodyStyle,
+  cardContainerStyle,
+  children,
+  childrenStyle,
+  icon,
+  iconBackgroundStyle,
```
spmcbride1201: To clarify, these are all blocks of inline styles that are used to override the default behavior of the `styled-components`, right? My expectation is that these allow customization, but aren't required to display a sane default.

src/components/Main/components/Settings/SettingsGrid.js
```diff
@@ -89,15 +93,32 @@ class SettingsGrid extends Component {
...
-                {
-                  title: "Cache Size",
-                  value: `${metricsCacheSize}`,
-                  children: button
-                }
-              ]}
+            <Card
+              title={"Cache Size"}
+              value={metricsCacheSize}
+              children={button}
```
spmcbride1201: Is there a reason that this can't be passed down like normal React component children?

```jsx
<Card>
  <Button/>
</Card>
```

src/components/Main/components/Settings/SettingsGrid.js
```diff
@@ -89,15 +93,32 @@ class SettingsGrid extends Component {
...
+              bodyStyle={{
+                fontSize: FONT_SIZE_HERO,
+                order: 1,
```
spmcbride1201: Reordering flex items feels a bit hacky to me. My expectation is that the consumer of this reusable component shouldn't have to think about flex object ordering

src/components/Main/components/Card/Card.js
karinakeen: Yes they are inline styles that will override the default styles.   Card has default styles set that you can view in Card stories.  We can add more styling to the default, if the we expect Card to look a certain way without overriding styles.

src/components/Main/components/Settings/SettingsGrid.js
karinakeen: Fixed alphabetization and passing of react component.    Custom flex object ordering is optional just like other inline styling, Card will display title body children in order.  For Metrics cache card, it displays body before title, so I specified them inline.

karinakeen:
This PR is WIP, pending further discussion.

PR #1199, opened 2017-12-11
fix: 1015 confirmation modal is pushed up too high in FF

### LINKED ISSUE #1015: FF 52, 56, 57: Modal pop up to reset the metrics cache is rendered off view port
## Expected Behavior

The pop up should pop up at the center of the view port

## Actual Behavior

Pops up at the top of the page, with only bottom part visible in the view port

  1. In FF 52,56 or 57 navigate to settings page and reset metrics cache

## Specifications

  - Version: 0.9.1
  - Platform: fabric

karinakeen:
Resolves #1015 

## Proposed Changes

  - assigned 100vh to confirmation modal container div created by React-Modal
  
## How to Test

  - Settings > click on `Clear Cache` button at the bottom and confirm that modal is not pushed up too high in all supported versions of Firefox browser


src/style/styleVariables.js
```diff
@@ -197,4 +197,15 @@ injectGlobal`
     text-decoration: none;
   }
 
+  /* fixes modal vertical alignment in firefox
```
7ruth: @karinakeen whats the rational for putting this css in the styleVariables vs styledModal.js file under the [class*="after-open"] section? Might be a better place for it since it would be right within the styled component and not depend on class name?

karinakeen: This scope applies to body tag, not to children of Modal.  When the modal is opened  ReactModal__Body--open class is added to the body tag that generates nested classnames that contain Modal, and this specific classname is the parent container of Modal tag.  This is one-off third party className, so I didn't create a separate third party injectGlobal block.

PR #1202, opened 2017-12-11
Fix: Table Latency 

### LINKED ISSUE #883: Stack Latency Values on top of each other
Currently our Go Routes, Go Functions, and JVM Routes tables display latency as two distinct columns:
image

These two columns should be stacked like in this mock.
image

Note: Disregard other UI discrepancies. They are out of scope for this PR

kaitmore:
Resolves #883 

## Proposed Changes

  -Added a TableColLatencyHeader functional component to wrap all the nested flexboxes


src/components/Main/components/Table/components/TableLineItem.js
```diff
@@ -150,11 +150,15 @@ export default class TableLineItem extends Component {
...
-        <TableCol style={{ textAlign: "right" }}>
-          {this.props.latency99}
+        <TableCol
+          style={{
+            textAlign: "right",
+            display: "flex",
+            flexDirection: "column"
+          }}
+        >
+          <div>{this.props.latency50}ms</div>
```
karinakeen: If you are styling the Latency header in this PR, you should make the percentages look smaller so they are not the same font-size as the word Latency.

PR #1207, opened 2017-12-12
feat: improved modal accessibility, complied and resolved react-modal…

### LINKED ISSUE #1121: Clear Cache modal, accessibility and focus improvements
## Expected Behavior

As focus shifts to this modal when you click into it, the outline should not be highlighted. 
As the focus shifts, the modal should be announced in the screen reader.

## Actual Behavior

The shift in focus is not announced in the voiceOver screen reader
Clicking with the mouse anywhere in the modal produces an outline around the whole modal

  1. Navigate to settings view and click clear cache button
  
## Specifications

  - Version: 1.0.1
  - Platform: fabric

7ruth:
… warning, adjusted modal screen positioning for mobile viewports

Resolves #1121 

## Proposed Changes

  -Added props to modal to make it more accessible with a screen reader (tested with mac's voiceOver)
  -Adjusted modal's positioning based on the view port size
  -Removed outline of the whole modal that was happening on click, while retaining outlines around buttons

## How to Test

  -Navigate to the settings page and click clear cache button


src/components/ConfirmationModal/ConfirmationModal.js
```diff
@@ -20,9 +20,16 @@ function ConfirmationModal({
   onConfirm,
   onCancel
 }) {
+  // The line below is necessary to designate a app element which will be given aria-visible false when modal is up so screen readers only focus on modal content
+  const rootAppElement = document.querySelectorAll("#root")[0];
   return (
     <StyledModal
       isOpen={isOpen}
+      aria={{
```
karinakeen: Great job adding aria labels to Modal.  I can hear the modal text in Chrome, Safari, Opera in voiceover.

src/components/ConfirmationModal/components/StyledModal.js
```diff
@@ -21,11 +21,21 @@ const StyledModal = styled(Modal)`
   opacity: 0;
   padding: ${spacingScale(3)};
   position: absolute;
-  top: 50vh;
+  top: 70vh;
   transform: translateX(-50%) translateY(50px) scale(0.98);
   transition: all 0.2s ease;
   width: ${modalWidthBasis};
 
+  @media only screen and (min-width: 320px) and (max-width: 767px) {
```
karinakeen: seems these breakpoints are constrained?  what happens if > 320 and < 1024px

karinakeen: add a case for < 1024

7ruth: While we can use 'responsive' to scale below 320px... I keep seeing 320px as the smallest dimension people design for. refs https://ux.stackexchange.com/questions/53911/what-is-the-smallest-screen-size-you-design-for && https://www.reddit.com/r/web_design/comments/3s5hnp/whats_the_smallest_cell_phone_width_for/

The case for above 1024 is the default already written, those media queries are like if statements for, what if view port is below 1024 <-> 768 and than again below 768 <-> 320 .

7ruth:
I also couldn't get FF to work with my Mac VoiceOver (its a FF + Mac VoiceOver issue though, https://www.applevis.com/forum/macos-mac-app-discussion/working-firefox-voiceover). In chrome & safari it picked up correct text for the modal. I assume Jaws would work fine with FF and most importantly pick up the the text in the modal.

PR #1208, opened 2017-12-12
Fix: Table interactions

### LINKED ISSUE #1180: Non-interactive table Items should not have hover state
## Expected Behavior
Items in a table that have no associated action shouldn't invite attempts at interaction.

## Actual Behavior
Threads table line items without stack trace data have a hover state indicating a click should provide some response, but clicking does nothing because there's no stack trace data.

  1. Navigate to the Threads page of a service instance
  2. Find a thread with no icon in the Trace column
  3. Hover the thread, observe pointer cursor and hover background color.

## Specifications

  - Version: 1.0.2
  - Platform: macOS
  - Subsystem: Safari 11

kaitmore:
Resolves #1180 

## Proposed Changes

  - Edited props in TableRow.js to make more semantic sense. `isOpen` now refers to the state of the drawer and `selectable` refers to whether the row has something for the user to interact with.
  - Moved ThreadsTable from index.js to ThreadsTable.js
 
  
## How to Test

  - Navigate to JVM => Threads. Confirm that rows without drawers (no stack trace) do not have a hover state and that the cursor is default. You can also double check other tables around the app if you feel so inclined.


src/components/Main/scenes/InstanceView/scenes/JVMInstanceView/scenes/Threads/components/ThreadsTable/index.js
```diff
@@ -1,95 +1,2 @@
...
-    case "TERMINATED":
-    case "BLOCKED":
-    case "NEW":
-      return "Stopped";
-    case "None":
-    default:
-      return "none";
-  }
-}
+import ThreadsTable from "./ThreadsTable";
```
spmcbride1201: Thanks for refactoring to follow our new convention!

src/components/Main/scenes/InstanceView/scenes/JVMInstanceView/scenes/Threads/components/ThreadsTable/components/ThreadsTableLineItem/index.js
```diff
@@ -84,7 +84,8 @@ export default class ThreadsTableLineItem extends Component {
 
     return (
       <TableRow
-        selectable={this.state.isOpen}
+        isOpen={this.state.isOpen}
```
spmcbride1201: Much cleaner API. Good work!

PR #1210, opened 2017-12-12
1204 buttonwrap breaks stylelint

spmcbride1201:
## Proposed Changes

  - Refactor ButtonWrap into smaller utility functions
  - Fix associated syntax errors
  - Add and update snapshots
  
Note: there still seems to be a syntax error breaking stylelint. I propose waiting for the resolution of https://github.com/styled-components/stylelint-processor-styled-components/issues/154 to be able to better understand why this is happening.

## How to Test

  - Run the app and click around
  - `npm test`
  - Visually inspect the snapshots, looking for major syntax errors


src/components/Button/components/ButtonWrap/ButtonWrap.js
```diff
@@ -0,0 +1,134 @@
...
+
+  &:focus,
+  &:active {
+    transition: all 0 ease;
+  }
+
+  &:not([disabled]) {
+    &:focus {
+      outline: none;
+      box-shadow: inset 0 0 0 rgba(255, 255, 255, 0.5),
```
7ruth: This box-shadow should be replace with ``` box-shadow: 0 0 0 2px #0aab2a; ``` which was deleted on line 673. Looks like this css prop never worked due to transparetize function not being specified so what we see on edge and what we are used to is box shadow that equal to 2px and this color #0aab2a;

src/components/Button/components/ButtonWrap/utils/generateButtonOrientation.js
```diff
@@ -0,0 +1,28 @@
+import { spacingScale } from "style/styleFunctions";
+
+// targetAlignment horizontal, vertical // default: horizontal;
```
7ruth: some comments over here

PR #1228, opened 2017-12-13
WIP: 108 filter out metrics that have values of only zero

### LINKED ISSUE #108: Provide ability to filter out metrics that have values of only 0 over the known period of time
Browsing ESS, there is currently around 1655 metrics expressed as timeseries data. Many of these metrics are just flat lines of zeros. The user should be able to filter out these values to focus on more meaningful metric.

image

karinakeen:
Resolves #108 

## Proposed Changes

  - add two checkboxes to filter metrics in Explorer view 
 - `Hide all metrics with only zero values`
 - `Hide all static metrics`
 - checking `Hide all static metrics` automatically check `Hide all metrics with only zero values` and disable it
 - add more unit tests to Explorer 
 - add styled components test
 - checkbox filters are router driven and retained along with search query on hard refresh
  - styling is TBD
## How to Test

  - Choose any Go/JVM Instance and navigate to Explorer View.  
  - check and uncheck both checkboxes to confirm metrics are filtered appropriately
  - run npm t


src/components/Main/components/Explorer/Explorer.js
```diff
@@ -40,7 +40,8 @@ class Explorer extends Component {
     super(props);
     this.state = {
       searchQuery: "",
-      lastPushedQueryString: ""
+      lastPushedQueryString: "",
+      isChecked: false
```
spmcbride1201: I think this should be a bit more descriptive, like `hideOnlyZeroMetrics`

src/components/Main/components/Explorer/Explorer.js
```diff
@@ -166,18 +175,27 @@ class Explorer extends Component {
     const selectedMetric = query.selectedMetric
       ? query.selectedMetric.replace(/%2F/g, "/")
       : undefined;
-    const { searchQuery } = this.state;
+    const { searchQuery, isChecked } = this.state;
+    let filteredKeys = keys;
+    // if a key metric does not have every data value at zero, display the key
+    if (isChecked) {
+      filteredKeys = filteredKeys.filter(key => {
```
spmcbride1201: This logic looks good, but it probably should be broken into a separate function to be more testable. I'd also expect to see a few unit tests on the function checking to make it's not accidentally over-filtering or under-filtering.

src/components/Main/components/Explorer/components/Inspector/components/InspectorSearch.js
```diff
@@ -42,4 +42,8 @@ const InspectorSearch = styled.input`
   }
 `;
 
+export const InspectorFilter = styled.input`
```
spmcbride1201: I'm insane, but please break this out into a separate named file and add a snapshot test.

src/components/Main/components/Explorer/Explorer.js
```diff
@@ -119,6 +120,14 @@ class Explorer extends Component {
     });
   };
 
+  /**
+   * onChange function invoked by Inspector when user checks box 'filter no metrics data' in the search bar
+   * @param {string}
+   */
+  onChange = checked => {
+    this.setState({ isChecked: !this.state.isChecked });
```
spmcbride1201: We need to make the state for this toggle router-driven too. Currently a hard refresh retains the search term and the selected item, but the toggle goes unchecked.

PR #1236, opened 2017-12-14
47 screen reader ux for graphs

### LINKED ISSUE #47: Screen Reader UX for graph components
The charts are currently incomprehensible to screen readers. It seems like the best UX for visually fully impaired folks would be presenting the data powering the chart as a table. However, some sort of a description of the statistics (mean, peak, trough) might be better still.

7ruth:
Resolves #47 

## Proposed Changes

  - Added an aria-label to line chart container.

Initially I added an offscreen div, as I remember that being the preferred method of displaying extensive amount of information that should be presented to screen readers but not sighted users. 

I don't exactly remember why it was preference method over aria-label or some other way, so in this team's eyes I couldn't whole-heartedly defend my decision other than "thats what we did at the accessibility consulting firm", thus I went with a simple aria-label, which works fine in Chrome and Safari with voiceOver (I would like to remind reviewers that voiceOver doesn't work with FF so I didn't test in it). 

There was some support that I found online for using offscreen divs:

''While you could certainly do this, the main upside to using screen reader only text over an aria-label is that the screen reader is not required to support ARIA, it’s simply reading text on the page that sighted users cannot see. While most screen readers support ARIA, you may not want to rely on it if you don’t have to." 
- https://accessible360.com/accessible360-blog/use-aria-label-screen-reader-text/

Read comments to accepted answer, looks like off-screen text is more reliable across all screenreaders than aria-label. 
- https://stackoverflow.com/questions/26032089/in-html-how-can-i-have-text-that-is-only-accessible-for-screen-readers-i-e-fo
 
https://webaim.org/techniques/css/invisiblecontent/

## How to Test

  -Navigate to Routes page, open drawer, turn on screen reader (cmd+f5) and hear all the glory


src/components/Main/components/GMLineChart/GMLineChart.js
```diff
@@ -43,10 +43,96 @@ export default function GMLineChart({
   timeSeries,
   title
 }) {
+  // distills timeSeries array that has time stamp to an array of values
+  let numericalTimeSeriesFunc = array => array[0].map(cell => cell[1]);
+  const numericalTimeSeries = numericalTimeSeriesFunc(timeSeries);
+
+  let median = array => {
```
karinakeen: Line 46-97 should be a separate function.

src/components/Main/components/GMLineChart/GMLineChart.js
```diff
@@ -43,10 +43,96 @@ export default function GMLineChart({
...
+      // if even, average of two middle numbers
+      median = (array[len / 2 - 1] + array[len / 2]) / 2;
+    } else {
+      // if odd, middle number only
+      median = array[(len - 1) / 2];
+    }
+    return median;
+  };
+
+  let average = array => array.reduce((a, b) => a + b) / array.length;
```
karinakeen: maybe /utils folder inside of GMLineChart to contain modularized functions

src/components/Main/components/Table/components/GMServiceTableLineItem.js
```diff
@@ -14,8 +14,20 @@ GMServiceTableLineItem.propTypes = {
 };
 
 function GMServiceTableLineItem({ status, name, uptime, path }) {
+  // this is done to search up the DOM tree to find table row and take away its focus to prevent outline on click while preserving tabbing outline
+  const blurTableRow = e => {
```
karinakeen: this should be a separate util function

src/components/Main/components/GMLineChart/GMLineChart.js
```diff
@@ -43,10 +43,96 @@ export default function GMLineChart({
...
+
+  let modes = array => {
+    let modes = [];
+    let count = [];
```
karinakeen: if you want to tighten some of these codes, could use combination of lodash functions such as _.maxBy and _.sortBy and _.meanBy

src/components/Main/components/GMLineChart/GMLineChart.js
```diff
@@ -43,10 +43,96 @@ export default function GMLineChart({
...
+  };
+
+  const screenReaderGraphDescription = `
+    A tabular representation of the ${title} chart data: 
+    median ${screenReaderGraphData.median}
+    average ${screenReaderGraphData.average}
+    mode ${
+      typeof screenReaderGraphData.mode !== "string" &&
+      screenReaderGraphData.mode.length > 1
+        ? "Mode has frequence of " +
```
karinakeen: frequency

src/components/Main/components/GMLineChart/GMLineChart.js
karinakeen: getting TypeError while clicking around
screen shot 2017-12-14 at 3 50 46 pm

src/components/Main/components/Table/components/GMServiceTableLineItem.js
```diff
@@ -14,8 +14,20 @@ GMServiceTableLineItem.propTypes = {
...
+    let node = e.target;
+    while (
+      typeof node.className !== "string" ||
```
karinakeen: I am not exactly sure what this function does.  Are you applying blur() to direct parent of clicked TableRow?

src/components/Main/components/GMLineChart/GMLineChart.js
karinakeen: I think there should be an short default message when the chart doesn't have any meaningful data, e.g. when all data values are zero
screen shot 2017-12-14 at 4 01 32 pm

karinakeen:
I can hear the chart script in Chrome and Safari, can't get it to work in other browsers (expected).  Please see other comments for suggested changes.

src/components/Main/components/Table/components/GMServiceTableLineItem.js
7ruth: Blur is applied to whichever child of tableRow has been clicked. That is what gets rid of blue outline on click while preserving it for keyboard navigation.

karinakeen: Probably better to do this in styled component with pseudo class then

src/components/Main/components/GMLineChart/GMLineChart.js
7ruth: So after taking a careful look, it looks like I can only use _.mean ...which replaces my one line average formula. I know that importing "_" has adverse effects on bundle size, thus I rather keep the things as they are. Other functions you suggested are heavier implementations of what I already is done in a simpler way aka, min = first term of sorter array.

src/components/Main/components/Table/components/GMServiceTableLineItem.js
7ruth: I am not sure what you mean by that... this deals with a scenario when you click on a row and you are clicking on an svg or something like that thats nester 4 layers deep into tableRow, resulting in tableRow being outlined. This finds tableRow and turns off focus so sighted users never see the outline while not disrupting keyboard use.

src/components/Main/components/Table/components/GMServiceTableLineItem.js
7ruth: I'll open a separate issue as there are a few of these function in the code base.

src/components/Main/components/GMLineChart/utils.test.js
```diff
@@ -0,0 +1,65 @@
...
+    expect(modes(mockTimeSeriesArrayNoMode)).toEqual(expectedOutputNoMode);
+  });
+
+  test("average provides a mean", () => {
+    const mockTimeSeriesArray = [1, 2, 3];
+
+    expect(average(mockTimeSeriesArray)).toBe(2);
+  });
+
+  xtest("screenReaderGraphDescription provides correct output ", () => {
```
karinakeen: Good job adding tests to your utils functions.  Is there reason why this test is being skipped?  Seems to throw a TypeError when I tried to run it.
screen shot 2017-12-18 at 6 30 24 pm

src/components/Main/components/Table/components/GMServiceTableLineItem.js
karinakeen: 👍

src/components/Main/components/Table/components/GMServiceTableLineItem.js
karinakeen: Confirmed this gets rid of extra highlight on the tablerow with mouseclick. 👍

src/components/Main/components/GMLineChart/utils.test.js
7ruth: @karinakeen should be working as expected now!

PR #1237, opened 2017-12-14
Feat: Latency tooltip

### LINKED ISSUE #1128: Tooltip info need for Latency table headings in Routes
## Expected Behavior
For sake of argument, I don't necessarily know what "Latency 50%" means. I expect to be able to hover the table heading and see a tooltip explaining precisely what it means.

## Actual Behavior
Hovering the table heading shows no additional info

## Specifications

  - Version: 1.0.1
  - Platform: macOS
  - Subsystem: Safari

kaitmore:
Resolves #1128 

## Proposed Changes

  - Implemented new tooltip component in table

## How to Test

  - Navigate to a route or function table and hover over the "Latency" table header.


karinakeen:
screen shot 2017-12-18 at 6 56 55 pm
Green underline visible in multiple browsers 
screen shot 2017-12-18 at 6 57 37 pm

kaitmore:
@karinakeen Yeah, I created that underline on purpose, usually tooltips have some sort of indication that you can hover over them. I am waiting for a review from @shanberg so he can make the final call on styling.

shanberg:
I agree we should have an underline, at least until the point where most of the interface has tooltips.

For styling, I think the green dotted may draw too much attention. In this specific tooltip, we're providing a help to folks who are curious or confused about latency, and I think it’s safe to assume there's a high probability that a user will either look closer at the label and notice a lighter underline, or hover the label and discover the tooltip. I recommend a lighter indicator, like a solid black line at 15% opacity.

On the tooltip itself, I think left-aligning the content will make it slightly easier to read long content, and shouldn't cause issues with single-line content.

PR #1244, opened 2017-12-18
Fix: Factored out blurTableRow into util function

### LINKED ISSUE #1239: Factor blurTableRow functions into a utility function
## Expected Behavior

Works as expected, but its used in several places

## Actual Behavior

We should factor it into utility function that gets imported into those places.

## Specifications

  - Version: 1.0.3
  - Platform: fabric

kaitmore:
Resolves #1239


kaitmore:
Thanks for catching those errors! Should be fixed now. As for `manageFocus` in ServicesListItem, it seems different than the other cases and it's the only one of its kind, so I'm going to leave it as is for now.

PR #1267, opened 2017-12-20
Test: Glyph/index.js

kaitmore:


src/components/Glyphs/index.test.js
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+import Glyph from "./index";
+import Bell from "./Bell";
```
karinakeen: Following the established  pattern, this test file should probably be renamed to Glyph.test.js and the component file as Glyph.js, and have index.js default export Glyph.js

PR #1284, opened 2017-12-21
Test: Services

### LINKED ISSUE #1254: Add tests for src/services/metrics

kaitmore:
Resolves #1254 

## Proposed Changes
  - deleted /threads/utils.js file with duplicate `buildDiscoveryServiceInstanceThreadsEndpoint` function
  - added test for /metrics/utils.test.js


src/services/instance/threads/utils.js
```diff
@@ -1,11 +0,0 @@
-import { getState } from "jumpstate";
```
karinakeen: indeed a duplicate function

src/services/instance/metrics/utils.test.js
```diff
@@ -0,0 +1,51 @@
...
+import mockstate from "json/mockReduxState";
+
+jest.mock("jumpstate", () => {
+  return {
+    getState: () => mockstate
+  };
+});
+
+jest.mock("utils/head", () => {
+  return { getFabricServer: () => "http://localhost:1337" };
```
karinakeen: these jest mocks are so handy

Issue #1298, opened 2018-01-02
Improve Contributing Section using Brad Frost questionnaire


spmcbride1201:

spmcbride1201:
Duplicate of #1230

PR #1304, opened 2018-01-03
1298: Create Contribution Guide

spmcbride1201:
Resolves: #1298 

## To Test
Read the documentation and test that all links resolve as expected.


spmcbride1201:
@kaitmore - For the default props thing, I think we haven't yet clearly established a standard. I opened #1317 to come up with one and then add to the contribution guide.

PR #1320, opened 2018-01-08
1313 remove serverless code

### LINKED ISSUE #1313: Remove the ability to run the dashboard without a fabric server
## Current Behavior

The Application contains the following static configuration options as meta tags in `index.html`:
* fabricServer
* runtime
* serviceName
* metricsEndpoint
* threadsEndpoint

With the exception of `fabricServer`, all of these options are used to deploy the dashboard without a backend in an architecture where the UI directly polls and scrapes metrics and threads endpoints.

## Proposed Changes

* Remove the capability to run the dashboard without a fabricServer

spmcbride1201:
Resolves #1313 

## Proposed Changes

  - Remove logic and unit tests associated with running the dashboard without a fabric server
  - Remove mock servers used for running the dashboard without a fabric server
  - Remove the old localForage implementation which had been commented out, as the code has gotten too stale to be easily reused in the future.
  
## How to Test

  - Start App
  - Manually test all app functionality looking for regressions


json-mock/go/metricsWithFunctions.json
```diff
@@ -1,130 +0,0 @@
-{
```
kaitmore: This file is still being referenced in json-mock/discovery-service/sds.js, so I get an error when running the mock server

spmcbride1201: I had fixed it locally, but the changed weren't committed. Whoops!

PR #1360, opened 2018-01-16
feat: allow users to access service view for service with no metrics

### LINKED ISSUE #1344: Allow users to click through to the Service view when a microservice doesn't have metrics
## Expected Behavior

When a microservice does not have metrics, it shows on the Fabric page with the "no metrics" icon. However, a user is able to select the microservice to open it's Service page to view the instances IDs and uptime values associated with the service. Each instance displays the no metrics icon and is unselectable, showing the disabled pointer state, and displaying a tooltip explaining that the microservice does not offer metrics.

## Actual Behavior

Currently, when a microservice does not provide metrics, the user is unable to click from the Fabric View into the Service View. The service card or list item is disabled with the hover and tooltip behavior that should instead be on the child instance links on the Service Page.

kaitmore:
Resolves #1344 

## Proposed Changes

  - Passed the serviceIsMetered prop down from the top level router so that GMServiceTableLineItem.js can decide whether to disable the link
  - Added an icon and a tooltip to GMServiceTableLineItem.js
  - Customized Tooltip to allow for disabling and custom styling
  
## How to Test

  - From both cards and list view, navigate to Grace Hopper. You should see the unmetered icon next to the instance id as well as a tooltip on hover.
- Navigate to a metered service and confirm there is no tooltip or icon
- Navigate to: http://localhost:3000/#/Grace%C2%B7Hopper%C2%B7Battleship%C2%B7Control%C2%B7Program/1.0/ee0fa3669fea7e9a0adea649c46bca56 
       You should be redirected back to the service view and see a notification


src/components/Main/components/Table/components/GMServiceTableLineItem.js
```diff
@@ -27,9 +36,29 @@ function GMServiceTableLineItem({ status, name, uptime, path }) {
...
+        <Tooltip
+          position="left"
+          disabled={serviceIsMetered}
+          containerStyle={{ border: "none" }}
+          content={
+            "This microservice instance does not have metrics to display."
+          }
+        >
+          <GMLink
+            disabled={serviceIsMetered ? null : true}
```
spmcbride1201: Cleaner as `disabled={!serviceIsMetered}`

PR #1373, opened 2018-01-16
research with-url-state

kaitmore:
implemented this in FabricGrid view so you can take a look
pros: 
- deleted 100 lines of code

cons: 
- slow
- doesn't seem like there is a way to decide what should be in the url bar. it looks more cluttered now with the long query string.


spmcbride1201:
I hear your reservations, but to me, the FabricGrid looks way simpler now with all of the local state moved to the router. We lost like 120 lines of pretty complicated glue code. After updating the dependency to the latest version, I don't notice any particular slowdown either. I haven't done it yet, but I strongly suspect that we could clean up the query string code to only show on non-defaults like before. Once you pull my changes and reevaluate the speed stuff you were hitting to let me know if you it's still slow, I'm willing to dig into this further.

kaitmore:
Hah, they released 1.2.1 a day after I implemented this 😆  I'm no longer noticing any speed issues.

tilleryd:
I like this a lot and I think we should run with it.  The only improvement I can suggest is to reduce the amount of boilerplate by creating a new module that imports `withUrlState` and `history`.  That way engineers don't have to import both modules.

```
# components/applyUrlState
import { withUrlState } from "with-url-state";
import history from "AppHistory";

export default function(Component) {
  return withUrlState(history, props => ({}))(Component);
}
```

```
import applyUrlState from "components/applyUrlState";
...
export default applyUrlState(FabricGrid);
```

PR #1439, opened 2018-01-29
WIP: 1348 react router redux

### LINKED ISSUE #1348: What value do we get out of React Router Redux?
We currently have this in the project, but I'm unclear where or how we're using it. We should investigate our current use case and the tradeoffs of keeping it versus dumping it.

spmcbride1201:
Resolves #1348 

## Proposed Changes

  - Refactor app to not use react-router-redux
  - Refactor old non-sds behavior from AppHeader
  - Refactor out old unused props in Main
  
## How to Test

  - npm test to see if stuff blows up
  - Run app and click around to see if things work as expected.


src/components/Main/scenes/FabricView/components/FabricGrid/FabricGrid.js
```diff
@@ -152,4 +153,4 @@ class FabricGrid extends Component {
   }
 }
 
-export default withUrlState()(FabricGrid);
+export default withRouter(withUrlState()(FabricGrid));
```
kaitmore: Just curious, why do we need to add this? It seems like it works fine without.

spmcbride1201: Thanks for the heads up! There's likely no rational reason other than I saw it was using react-router-redux. Going to address in #1440

PR #1442, opened 2018-01-30
fix: instance polling error

### LINKED ISSUE #1365: Instance Polling Controls error when used before selecting an instance
## Expected Behavior

A user should be able to change the instance polling controls when an instance is not yet selected. The start / pause polling button should be disabled (because it has not meaning when an instance isn't selected), and the instance polling rate selector should be functional without causing an error.

## Actual Behavior

If a user sets the instance polling controls before an instance is selected, a confusing error message is presented

  1. Navigate to http://localhost:3000/#/settings
  2. Click Start Polling under Instance Polling and see error
  3. Drag the polling interval selector under Instance Polling and see error
  
image

## Specifications

  - Version: v1.0.3
  - Platform: Mac High Sierra
  - Subsystem: Chrome 63

kaitmore:
Resolves #1365 

## Proposed Changes

  - Added a bool `isDisabled` which is true when there is no selectedInstance. 
  - Added a tooltip to let the user know they need to select an instance
  - Added disabled styles to the polling button svg and instance slider
  
## How to Test

  - Navigate to http://localhost:3000/#/settings
  - Notice that the button is disabled and a tooltip renders on hover. The slider can still move but does not trigger any actions.
  - Go select an instance
  - Check that the button is enabled and you are able to set the polling interval as expected


kaitmore:
@spmcbride1201 I tried turning react-input-range into a styled component, but it just won't apply the classname selectors. I think this is why I implemented InputRange.js this way in the first place.

I changed this to use the react-input-range's disabled prop, and update styles to match.

kaitmore:
I think it's not possible to turn this into a styled component because InputRange doesn't pass through it's className props: 
https://github.com/styled-components/styled-components/issues/265#issuecomment-263321874
https://www.styled-components.com/docs/advanced#existing-css

spmcbride1201:
Opened https://github.com/DecipherNow/gm-fabric-dashboard/issues/1448 to table the keyboard focus issue until later.

PR #1447, opened 2018-01-31
Refactor App to use Slugs

### LINKED ISSUE #805: Sidebar and breadcrumb links to services go to the dashboard
When viewing a specific instance of a service, the link at the top the sidebar with the service name and the breadcrumb with the service name (immediately to the right of the "fabric" breadcrumb) both navigate back to the dashboard page rather than to the expected service page (the one containing the list of instances).

### LINKED ISSUE #1437: Persistent Parenthetical Problem
## Actual Behavior

When a microservice has a parenthesis in the name, the instance dashboard renders an error condition when an instance is selected.

  1.
  1.
  1.
  
## Specifications

  - Version:
  - Platform:
  - Subsystem:

spmcbride1201:
Note: Refactor currently breaks tons of tests. The mock JSON data needs to get refactored.

Resolves #805 
Resolves #1437 

## Proposed Changes

  - Generate slugs when we inject services from the discovery-service using the service name and version
  - Key services in Redux off the slug
  - Refactor Routers to collapse service name and version into this new slug
  
## How to Test

  - Run app and check for regressions
  - Run storybook and check for regressions


src/components/AppHeader/AppHeader.js
```diff
@@ -27,18 +27,17 @@ function AppHeader({ location: { pathname } }) {
 }
 
 function getTitle(pathname) {
-  const [root, version = "", instance = ""] = decodeParameter(pathname)
+  const [selectedServiceSlug = "", selectedInstanaceID = ""] = pathname
```
kaitmore: typo here **selectedInstanceID

We should probably just connect AppHeader to state so we can lookup the real service name using the path. Maybe a follow up issue?

src/components/Main/scenes/FabricView/FabricRouter.js
```diff
@@ -0,0 +1,198 @@
...
+              message = `${serviceSlug} does not have metrics to display`;
+              pathname = `/${serviceSlug}`;
+            }
+          }
+
+          // If the services object has not been passed to the router yet,
+          // or if the user is authorized and the service exists and is metered
+          // as determined by the message not getting set above
+          // then render the instance router, else redirect with error message
+          return servicesAreNotLoaded || !message ? (
```
kaitmore: this is a lot better 🙌

src/components/Main/scenes/FabricView/FabricRouter.js
```diff
@@ -0,0 +1,198 @@
...
+
+          // Set a message to pass to location state if one of the following checks fail
+          let message,
+            pathname = "/";
+
+          // Checks are ordered by priority of the message
+          // Check if the services are loaded, but the service slug didn't match
+          if (!servicesAreNotLoaded) {
+            if (!service) {
+              message = `${serviceSlug} is not a known microservice`;
```
kaitmore: Probably should use just `service` for messages instead of `serviceSlug`

kaitmore:
Awesome job!!! You've cleaned up an amazing amount of stuff here. 

I didn't notice any major regressions clicking through the app. The only thing is the service name in the header.

src/components/AppHeader/AppHeader.js
spmcbride1201: Yeah, I agree that we should pull the actual service name. I'll open an issue for it.

PR #1480, opened 2018-02-07
feat: table i18n

### LINKED ISSUE #1469: Add i18n to Table & TableToolbar

kaitmore:
Resolves #1469


src/components/Main/components/Table/Table.js
```diff
@@ -27,13 +27,7 @@ Table.propTypes = {
...
  * @returns
  */
-export default function Table({
-  items,
-  type,
-  selectedServiceSlug,
-  serviceIsMetered,
-  status
-}) {
+function Table({ items, type, selectedServiceSlug, serviceIsMetered, status }) {
```
spmcbride1201: Was this just an aesthetic change? This aligns our default exports with what we have to do for classes, so it might be better just to do that throughout the app. Was that your thought process?

src/components/Main/components/Table/components/TableLineItem.js
```diff
@@ -41,9 +42,10 @@ const FlexParent = styled.div`
  * @class TableLineItem
  * @extends {Component}
  */
-export default class TableLineItem extends Component {
+class TableLineItem extends Component {
   static propTypes = {
     errorPercent: PropTypes.string.isRequired,
+    intl: PropTypes.object.isRequired,
```
spmcbride1201: Can we build out a shape for this?

src/components/Main/components/Table/utils.js
```diff
@@ -10,25 +10,65 @@ import TableColLatencyHeader from "components/Main/components/TableColLatencyHea
...
+      />,
+      <FormattedMessage
+        id="table.uptime"
+        defaultMessage="Uptime"
+        description="Service view table header title"
+      />
+    ];
+  } else {
+    headerNames = [
+      <FormattedMessage
```
spmcbride1201: Can you please open an issue for this and tag it with a question like "How do we handle dynamic generation of strings that we want to internationalize?" I figure we can tag other strings like this and revisit the issue in a future sprint to make sure our architecture makes sense.

src/components/Main/components/Table/Table.js
kaitmore: Woops... this was just left over by mistake. I was going to wrap Table with injectIntl() so I could pass `intl` to the getTableHeaders function, but then I realized I didn't need to.

PR #1484, opened 2018-02-07
refactor: appheader

### LINKED ISSUE #1450: Refactor AppHeader to render the name, not the slug

  1.
  1.
  1.
  
## Specifications

  - Version:
  - Platform:
  - Subsystem:

kaitmore:
Resolves #1450 

## Proposed Changes

  - refactored the getTitle function to lookup service name from slug
  - refactored breadcrumbs. it needed both the service name and the slug, so I passed it a serviceNameLookup object


src/components/AppHeader/AppHeader.js
```diff
@@ -10,37 +12,49 @@ import { trimID } from "utils";
...
-  // Else just display Fabric
-  if (selectedInstanceID) {
-    return `${selectedServiceSlug} : ${trimID(selectedInstanceID)}`;
-  } else if (selectedServiceSlug) {
-    return selectedServiceSlug;
+function getTitle(pathArray, services) {
+  const [selectedServiceSlug = "", selectedInstance = ""] = pathArray;
+
+  const selectedService =
+    (services[selectedServiceSlug] && services[selectedServiceSlug].name) ||
```
spmcbride1201: We should probably add the version to the end of the name string to make that more explicit.

PR #1488, opened 2018-02-08
refactor: change state management to better render units dynamically

### LINKED ISSUE #1478: Graph labels have some confusing labels
## Actual Behavior
 The sub-panes in the service instance views have some confusing labels.  Initially I thought it would be Kb, Mb, Gb, but it seems to be 1000, Million, Billion (of bytes presumably?)

A label specifying/explaining the units, or perhaps changing to Kb, Mb, Gb would help clarify.

  1.
  1.
  1.
  
screen shot 2018-02-06 at 10 58 50 am

## Specifications
  - Version: 1.0.4
  - Platform: Chrome
  - Subsystem:

spmcbride1201:
Resolves #1478 

## Things I may need to do
* I noticed that it seems like the lifecycle hooks are a bit messed up when I messed around in storybook. Something is probably off with how I'm storing local state and comparing things. 
* The Y axis also really should render the correct units
* I also need some realistic unit test around the new utility functions
  
## How to Test

  - Click around the app and look for regressions. Notice how the unit labels now render
  - Carefully review my crazy change


src/utils/index.js
```diff
@@ -119,6 +120,48 @@ export function formatAsDecimalString(number, numberOfDecimals = 3) {
...
+ * and rounding to a desired unit of precision
+ * @export
+ * @param {number} rawValue
+ * @param {string} baseUnit
+ * @param {string} resultUnit
+ * @param {number} precision
+ * @returns string
+ */
+export function formatMetricString(rawValue, baseUnit, resultUnit, precision) {
+  precisionValueGenerator(precision);
```
kaitmore: Seems like this might be leftover^^

PR #1503, opened 2018-02-12
chore: set CircleCI timezone to EST before running tests

### LINKED ISSUE #1483: Configure CircleCI to use EST when running our unit tests to prevent timezone issues from breakout our snapshots that use shallow()
JVMHeaderContent, GoHeaderContent, and Summary views are all using shallow instead of render because of issues with using Date.now in the Uptime component.

When doing a full render, CI fails because dates/times are based on the timezone of the CircleCI server https://medium.com/front-end-hacking/jest-snapshot-testing-with-dates-and-times-f3badb8f1d87

This solution uses `moment`, which is a big library and would require refactoring the Uptime component just so our snapshots work. There is probably a better solution out there like setting the timezone in our circleci config.

spmcbride1201:
Resolves #1483 

## Proposed Changes

  - Set timezone in CircleCI Config
  - Also explicitly set the TZ environment variable immediately before running `npm test` to force POSIX behavior between Mac and Linux (this sacrifices clean Windows support pending use of `moment` or `date-fns`
  
## How to Test

  - Confirm that `npm test` passes locally without snapshot errors
  - Confirm that CircleCI passed without snapshot errors
  - `npm start` and confirm the Summary components don't have visual regressions


spmcbride1201:
Still getting really odd errors after setting the timezone to NYC time: 
image

spmcbride1201:
I'm not totally clear why this was the case, but it seems that the timezone information when we locally run `npm test` hasn't been correct. Mac OS also seems to express time zones slightly differently. By setting the timezone environment variable before running npm test, we force things to behave the same on Mac and Linux. I'm going to remove the timezone from the UI now that I'm done testing, but this should be ready for review in a few minutes.

spmcbride1201:
@kaitmore - I addressed your comments!

PR #1507, opened 2018-02-13
build: add git safety belts prior to publishing

### LINKED ISSUE #1349: Publish Script should Error out if there are uncommitted file changes
## Proposed Changes

* Modify the publish script to make sure that the project doesn't have any uncommitted file changes and bail out if there are. This would help protect against builds with accidental content from getting published to Dockerhub

spmcbride1201:
Resolves #1349

## Proposed Changes

  - Add safety checks to our publish script to be sure that we have master checked out and that there aren't any local uncommitted changes
  
## How to Test

  - Run `npm run publish -- --dryRun` and confirm that the dry run works as before
  - Run `npm run publish` and confirm that it errors out for not having master checked out
  - Change my test to resolve the promise if we are not on master
  - Run `npm run publish` and confirm that it errors out for having local uncommitted changes
  - Locally commit your changes (being sure not to push them to origin) and run `npm test` to make sure you get to the first confirmation question. Hit ctrl+C to exit without taking any action


kaitmore:
I assume you meant `npm run publish` not `npm test`

Everything is working for me except for the last step. After locally committing, `confirmNoUncommited` is still failing (which is to be expected since it's a dry run) but it's saying I having 2 uncommitted changes.

PR #1541, opened 2018-02-21
test: base instance view

kaitmore:
Resolves #

## Proposed Changes

  - Added a helper function that asserts a testcafe array-like thingy (?) is in order
  - Added a test for views common to both JVM & GO. Since Routes and Functions views are functionally identical, we decided that it was ok to just test Routes view to avoid duplication.
  - Updated the `test:e2e:local` command to use headless chrome
## How to Test

  - `npm run test:e2e:local`


e2e-tests/tests/base-instance-view.js
```diff
@@ -0,0 +1,240 @@
...
+  // Average Response Time readout
+  const summaryAvgResponseTimeReadoutText = await instanceViewModel.summaryAvgResponseTimeReadout.find(
+    "span"
+  ).innerText;
+
+  await t
+    .expect(instanceViewModel.summaryAvgResponseTimeReadout)
+    .ok()
+    .expect(summaryAvgResponseTimeReadoutText)
+    .match(/[+-]?([0-9]*[.])?[0-9]+\s(ms)|-/); // match `<float> ms` or a dash -
```
spmcbride1201: For regular expressions, I now like to create and save an example with good sample data on regexr so folks can get a grasp of how the expression behaves. An example of this is https://regexr.com/3kof2. Can you please do the same?

e2e-tests/tests/base-instance-view.js
```diff
@@ -0,0 +1,240 @@
...
+const instanceViewModel = new BaseInstanceViewModel();
+const fabricViewModel = new FabricViewModel();
+const serviceViewModel = new ServiceViewModel();
+
+// The following test all views common to both JVM and GO instance views
+fixture`Instance View`.page`http://localhost:3000`.beforeEach(
+  async t =>
+    await t
+      .click(fabricViewModel.linkStable)
+      .click(fabricViewModel.servicesCards.nth(0))
```
jsweetland: Same comment as the other PR, might want to make sure at least one card is there before attempting to click it. Same goes for the line immediately after this one.

e2e-tests/tests/base-instance-view.js
```diff
@@ -0,0 +1,240 @@
...
+test("Verify Summary View Layout", async t => {
+  //  Expect the 2 layout sections to be present
+  await t
+    .expect(instanceViewModel.summaryVitalsSection)
+    .ok()
+    .expect(instanceViewModel.summaryStatisticsSection)
+    .ok();
+
+  // Wait for readout cards to be populated
+  await t.wait(3000);
```
jsweetland: Is there something we can key on for this rather than waiting a static amount of time?

e2e-tests/tests/base-instance-view.js
```diff
@@ -0,0 +1,240 @@
...
+  let aRoutesTableRow = await instanceViewModel.routesTableRows.nth(-1);
+  await t.click(aRoutesTableRow);
+  aRoutesTableChart = await instanceViewModel.routesChart.count;
+  await t.expect(aRoutesTableChart).eql(1);
+
+  // Search
+  await t.typeText(instanceViewModel.inputSearchRoutes, "r");
+  routesTableRowsCount = await instanceViewModel.routesTableRows.count;
+  await t.expect(routesTableRowsCount).gt(0);
+  //  Clear search and expect to see all routes
```
jsweetland: Nitpick, add an EOL before this line to maintain consistent whitespace.

e2e-tests/tests/base-instance-view.js
```diff
@@ -0,0 +1,240 @@
...
+  await t
+    .click(instanceViewModel.selectSortRoutes)
+    .click(instanceViewModel.optionSortRoutesLatency99);
+
+  // Expect that the table rows are sorted in ascending order by latency 99%
+  await assertSorted(t, instanceViewModel.routesTableColsLatency99);
+});
+
+test("Verify Explorer View Functionality", async t => {
+  await t.wait(3000);
```
jsweetland: Same as above, can we key on something instead of waiting a static amount of time?

e2e-tests/tests/base-instance-view.js
```diff
@@ -0,0 +1,240 @@
...
+  await t.expect(inspectorGraphCount).eql(0);
+
+  // Click the first inspector item
+  const firstItem = await instanceViewModel.inspectorItems.nth(0);
+  const firstItemTitle = await instanceViewModel.inspectorItems.nth(0)
+    .innerText;
+  await t.click(firstItem);
+
+  // Wait for the graph to populate
+  await t.wait(1000);
```
jsweetland: Same as above

e2e-tests/tests/base-instance-view.js
```diff
@@ -0,0 +1,240 @@
...
+    .eql(firstItemGraphTitle);
+
+  // Grab the first inspector item text to use as a search string
+  const searchString = await instanceViewModel.inspectorItems.nth(0).innerText;
+
+  await t.typeText(instanceViewModel.inputSearchMetrics, searchString, {
+    paste: true
+  });
+
+  await t.wait(1000);
```
jsweetland: Same

PR #1560, opened 2018-02-23
test: add Spanish internationalization tests

### LINKED ISSUE #1520: Add Internationalization E2E Tests

jsweetland:
Resolves #1520 

## Proposed Changes

  - Makes view models language-sensitive
  - Adds internationalization tests for Spanish


e2e-tests/view-models/fabric-view-model.js
```diff
@@ -1,52 +1,79 @@
...
 
 export default class FabricViewModel extends BaseViewModel {
-  constructor() {
+  constructor(locale = "en-US") {
     super();
 
     // Tab bar
-    this.linkAllServices = ReactSelector("TabLink").withText("All Services");
+    this.linkAllServices = ReactSelector("TabLink").withText(
+      toTitleCase(messages[locale]["fabricHeaderContent"]["allServices"])
```
kaitmore: There is a lodash function that does this, might be worth it to just use that instead of creating the extra helper function. 

You can `import _ from 'lodash'` and then use it like
```javascript
_.startCase(messages[locale]["fabricHeaderContent"]["allServices"])
```

PR #1561, opened 2018-02-23
fix: resolve issues with safari/edge/ie

kaitmore:
## Proposed Changes

  - While watching one of the browserstack videos I noticed some shenanigans with the hash in the url bar, and then found this related testcafe issue: https://github.com/DevExpress/testcafe/issues/2005
Adding trailing slashes to all our page fixtures solved it!
  - I also fixed the instance view (go) test that was failing yesterday due to some missing selectors
  - Enabled `test:e2e` step in circle ci 

## How to Test

  - `npm run test:e2e` the tests are really slow, but they should pass!


e2e-tests/view-models/instance-go-view-model.js
```diff
@@ -5,6 +5,13 @@ import BaseInstanceViewModel from "./base-instance-view-model";
 export default class InstanceGoViewModel extends BaseInstanceViewModel {
   constructor() {
     super();
+    // Summary
+    this.summaryHostCPUReadout = ReactSelector(
+      "ReadoutDisplay ItemDisplay"
+    ).withText("Host CPU Utilized");
+    this.summaryMemoryReadout = ReactSelector(
+      "ReadoutDisplay ItemDisplay"
+    ).withText("Memory Utilized");
```
jsweetland: This is good for this PR but will need to be updated to follow the same scheme as the other UI text stuff for internationalization.

PR #1570, opened 2018-02-27
feat: improved app header style

shanberg:
## Proposed Changes

  - Cleaner style for App header.
  - Improved margins for page title text.
  - Better style for background graphic


src/components/AppHeader/components/AppHeaderContainer.js
```diff
@@ -1,25 +1,24 @@
 import styled from "styled-components";
 
 import { COLOR_ALT_BACKGROUND } from "style/styleVariables";
-
-import BannerBackgroundImage from "images/mesh-long.png";
+import BannerBackgroundImage from "images/mesh-4k.png";
```
spmcbride1201: LOL. A 5.8mb image. I ran this through an optimizer and got it down to 1.2mb, but that's still pretty expensive. I'd like for us to compare this to using an SVG here. If an SVG isn't suitable, we need to use `srcset`, provide a few different sizes, and let the browser figure it out. https://css-tricks.com/responsive-images-css/. Also https://www.youtube.com/watch?v=eMtcF7f9PUU. I think using `image-set` is okay. All the images in the image set should be optimized.

spmcbride1201: If you like that youtube, you might like the course it's from: https://www.udacity.com/course/responsive-images--ud882

shanberg:
Adding that image was a mistake. I had the size down to 180kb. Will fix.

shanberg:
Size reduced to 100k.

